PR #4, opened 2014-07-26
Bugfix/issue1948: make os.stat behave like CPython (raise OSError on empty path)

hjacobs:
Fix for http://bugs.jython.org/issue1948


jimbaker:
I fixed this another way - with a lot more changes in how we do stat - but thanks for taking a look!

PR #7, opened 2014-08-04
To enable %f in datetime.strptime to fix issue# 2166

ewibowo:
- Modified Lib/datetime.strptime to call _strtime._strptime_datetime that accepts %f option
- Added method _strptime_datetime to lib-python/2.7/_strptime.py to extend to accept % f option


pekkaklarck:
This is the original bug report: http://bugs.jython.org/issue2166

jimbaker:
Thanks for taking a look. I will see how we can apply this change, but using the approach we do in Jython, as summarized in Jeff's comments.

jimbaker:
We updated Lib/datetime.py to the latest version from PyPy as of https://hg.python.org/jython/rev/daa6bf9a14d5, which resolved this and other issues. Thanks for taking a look at this problem, it's always very helpful!

PR #12, opened 2014-10-16
Implement Popen.pid (http://bugs.jython.org/issue2221)

pekkaklarck:
This code has been tested on Linux and Windows 7, and same approach was
tested earlier also on OSX.

If underlying Process implementation is not what we expect (i.e. doesn't
have 'pid' or 'handle' field, depending on OS), pid will be None. This
functionality is tested by querying a non-existing field.

Pid would also be None if a security manager would restricts reading
aforementioned private fields. In practice execution stops already in
'_setup_env' under such security manager, though.


pekkaklarck:
This PR has been merged manually to https://hg.python.org/jython/ by @agronholm.

PR #14, opened 2015-01-23
Fix Popen.kill on POSIX and Popen.send_signal in general.

pekkaklarck:
This PR fixes two highly related issues:

1. `Popen.send_signal` now works instead of failing for `AttributeError` due to `signal.CTRL_C_EVENT` not being available. Reported as http://bugs.jython.org/issue2219.

  On Windows only `SIGTERM` is supported by mapping it to Popen.terminate() exactly like CPython does. `CTRL_C_EVENT` and `CTRL_BREAK_EVENT` are not supported.

  Elsewhere all signals ought to be supported. This assumes signals are supported in general by Jython on that platform and `Popen.pid` is got correctly.

2. `Popen.kill` now forcefully kills the process also outside Windows. Reported http://bugs.jython.org/issue2220.

  Implementation uses `Process.destroyForcibly()` when available (new in Java 8) and `Popen.send_signal(signal.SIGKILL)` otherwise.

### DIFF
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -1425,24 +1425,36 @@ class Popen(object):
                     self._stdin_thread.interrupt()
             return self.returncode
 
-        def send_signal(self, sig):
-            """Send a signal to the process
-            """
-            if sig == signal.SIGTERM:
-                self.terminate()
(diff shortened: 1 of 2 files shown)


jimbaker:
+1, LGTM!

PR #15, opened 2015-02-07
Restore get_line_buffer method in readline.py

dieterv77:
Seems to have been removed with commit 56eb138d330762d25ceabffd1ee04342b186f071


jimbaker:
+1, makes sense to me

jimbaker:
Merged

PR #16, opened 2015-02-18
Update external postgres jar

dieterv77:
I was having problems with an application of mine that uses postgres, becausethe jar that ships with jython is too old to work with my app.  I figured i could just update the one in jython.

Not sure if this is how this is usually handled.  Feel free to decline the pull request if there's a better way.  Obviously i can override the jar locally if i need to.


jimbaker:
Should finally do this update. We were planning to move to Gradle, but that's a 2.7.2 thing at this point.

jimbaker:
Closing this out, since jar updates is really done as a normal matter of maintenance.

PR #17, opened 2015-03-03
PyUnicode cannot be serialized

orius123:
since beta 4 PyUnicode cannot be serialized because IndexTranslator was added and IndexTranslator doens't implement serialiazble.

### DIFF
--- a/Lib/test/test_java_integration.py
+++ b/Lib/test/test_java_integration.py
@@ -644,6 +644,10 @@ class SerializationTest(unittest.TestCase):
         date_list = [Date(), Date()]
         self.assertEqual(date_list, roundtrip_serialization(date_list))
 
+    def test_java_serialization_unicode(self):
+            str = "test_str"
+            self.assertEqual(str, roundtrip_serialization(str))
+
     def test_java_serialization_pycode(self):
 
         def universal_answer():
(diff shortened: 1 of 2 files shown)


orius123:
thi is actually blocking my porject from upgrading to latest beta..

jimbaker:
+1, looks reasonable. I'm glad you filed the bug in bugs.jython.org, because that's where we are tracking prioritization.

jimbaker:
The proposed test for test_java_integration didn't actually verify the fix, but I did add a new JUnit test to do so, see https://hg.python.org/jython/rev/8e85dc69027c. Thanks for finding this bug and its solution, now merged in.

orius123:
Thanks for the fast response!!                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        From: Jim BakerSent: Friday, 13 March 2015 20:47To: jythontools/jythonReply To: jythontools/jythonCc: Eliya SadanSubject: Re: [jython] PyUnicode cannot be serialized (#17)The proposed test for test_java_integration didn't actually verify the fix, but I did add a new JUnit test to do so, see https://hg.python.org/jython/rev/8e85dc69027c. Thanks for finding this bug and its solution, now merged in.

—Reply to this email directly or view it on GitHub.

PR #18, opened 2015-03-04
Coerce anything implementing the iteration protocol to Java arrays

ericfrederich:
Specifically this fixes issues with Jython getting "netref" objects
from RPyC that reference lists but don't explicicly subclass them.

Now anything that implements __len__ and __getitem__ can be coerced


ericfrederich:
Damn, I saw 2.7 was released and notices this code wasn't merged :(

ericfrederich:
Damn, I saw 2.7 was released and notices this code wasn't merged :(

darjus:
Hey Eric,

Thanks for reporting this and sorry we missed it. Can you also add a test for this issue?

ericfrederich:
@darjus sure, I'll try to find some time to re-visit this and add an appropriate test.

ericfrederich:
I'm not really sure where to put the test.  The test is a Python script but it depends on a Java class.

With this patch it runs without exceptions; without it you get an error about not being able to coerce.
Here are the two files.  I'd appreciate it if someone could just add this test for me, I'd need some guidance to do it myself.

```Java
public class Foo {
    public static void Bar(String[] args) {
        for (String s: args) {
            System.out.println("GOT " + s);
        }
    }    
}
```

```Python
#!/usr/bin/env jython

from collections import Sequence
import Foo

class MyType(Sequence):
    def __init__(self, l):
        self.l = l
    def __getitem__(self, i):
        return self.l[i]
    def __len__(self):
        return len(self.l)

Foo.Bar(['one', 'two'])
Foo.Bar(MyType(['three', 'four']))
```

ericfrederich:
FYI: I just decided to do a rebase rather than a merge to keep this pull request more simple.
Sorry, I know its bad practice to do non-fast-forward things on remotes.

Stewori:
@Eric
I once had the same question regarding GC-stuff. For some tests I needed `GCHelper.class`. It appeared that there are concurrent paradigms around where to put it and I failed to obtain a clear answer which one to use:

1) If you look into the folder `jython/Lib/test`, where Jython-specific tests should house, there are several .jar-files around, some of them containing class-files which I assume are dependencies for some tests.

2) The build-process includes files in `jython/tests/java/javatests` into `jython.jar/javatests`. In your .py test-file you can import the class via `from javatests import whatever`

For `GCHelper.class` I chose option (2) because I felt that it is somehow cleaner as it does not contaminate the Lib-folder with Java-files, which would be contradictory to the goal to merge jython/Lib into CPython's original standard-lib one day. For the same reason I would like to encourage writing Jython-specific tests such that they would not fail on CPython (but are skipped instead). However this is no strict requirement though. To achive this, use a decorator like
```python
@unittest.skipUnless(test_support.is_jython,
        '''
        The test involves Jython-specifics and is thus not supported by
        non-Jython interpreters.
        ''')
```

Also, do the import of your Java-class test-helper either in the skipUnless-annotated code or, if you use it frequently, surround it by try/catch. Example from `test_gc_jy.py`:
```python
try:
    from java.lang import System, Runnable, Class
    from javatests import GCTestHelper
except ImportError:
    #i.e. not Jython is running
    pass
```

ericfrederich:
Cool... thanks for the help @Stewori .
I added a test for it.

ericfrederich:
@darjus I added a test for this (says 27 days ago).  Wondering if this can get merged.
I think this is what the behavior of Jython should be.

ericfrederich:
Bump...
Any chance of this getting merged?

darjus:
Hey Eric, sorry for the long delay on this, it looks good! One more thing i need you to do tough is to sign contributor agreements: https://www.python.org/psf/contrib/contrib-form-jython/ https://www.python.org/psf/contrib/contrib-form/

Let me know if you already did that.

ericfrederich:
Yeah, it sounded familiar.  Searched my email and see I signed it July 22.

darjus:
Pulled!

PR #27, opened 2015-12-08
Set direct mode if sys.exectuable is None or empty for standalone installations

teeohhem:
Otherwise, calls to setup() will eventually fail because sys.exectuable is None and the temporary script generated in byte_compile will fail.

Also, in the commands/build.py, there is also an assumption that sys.executable will never be None, which leads to ```os.path.normpath(sys.executable)``` causing string errors. Looks like this bug was introduced in: https://mail.python.org/pipermail/jython-checkins/2015-September/001547.html

### DIFF
--- a/Lib/distutils/command/build.py
+++ b/Lib/distutils/command/build.py
@@ -114,7 +114,7 @@ class build(Command):
             self.build_scripts = os.path.join(self.build_base,
                                               'scripts-' + sys.version[0:3])
 
-        if self.executable is None:
+        if self.executable is None and sys.executable:  # sys.executable is None for standalone jar installs...be careful
             self.executable = os.path.normpath(sys.executable)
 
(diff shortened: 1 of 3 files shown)


darjus:
@jimbaker can you take a quick look at this since I'm not very familiar with this code? Thanks!
And thanks @teeohhem for the PR!

teeohhem:
I added a test, but I am unsure of two things:
1) Whether this file was the right location for the test and
2) This test was always grabbing my system's version of the distutil.command.build lib, and not Jython's Lib version.

I was able to verify the test through other means, but I want to make sure this is going to work inside of Jython.

jimbaker:
Closing in favor of what was done in https://hg.python.org/jython/rev/d9c1b18fcb9b

PR #28, opened 2016-01-12
zipimporter supports multi-byte characters in pathnames

agaman:
I have a JAR which contains compiled Scala classes as well as Jython. The [shapeless library](https://github.com/milessabin/shapeless) contains a class that ends up as a class file in the JAR named "$tilde$qmark$greater$λ.class" (notice the lambda).

Source here: [https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/package.scala](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/package.scala)

zipimporter eagerly transforms all paths in the JAR into PyStrings as it iterates over them, but when it hits this file with a multi-byte character in the name, the PyString constructor throws
`IllegalArgumentException("Cannot create PyString with non-byte value");`

Replacing the usage of `new PyString()` with `Py.newStringOrUnicode()` appears to solve the issue.


jimbaker:
Looks like a reasonable solution, but your test does not actually work, as seen in the the Travis CI report. Also the test should be in test_zipimport_jy, which we use to separate out tests we have specifically done for Jython (and they need not not necessarily about Java integration).

Procedural steps:
1. Please submit the contributor agreement: https://www.python.org/psf/contrib/contrib-form/
2. Submit a corresponding bug in http://bugs.jython.org/, just so we can track

Thanks!

jimbaker:
@agaman Any updates on my message from a few weeks ago?

Stewori:
I'd like to merge this for 2.7.1 without the test-code, given Jim's concerns about it.
IMO switching to Py.newStringOrUnicode() isn't a dangerous code-change and I already asserted that it doesn't break test_zipimport, test_zipimport_support or any other regrtest.
@agaman it would be ideal if you could sign https://www.python.org/psf/contrib/contrib-form-jython/, however I guess given that the code-change is only two lines we can accept it already without that.

Stewori:
Merged (without the test) as of https://github.com/jythontools/jython/commit/f7c73f4525313f6c1397c03e08b7ebd622f428e7

Can be closed.

@agaman It would be ideal if you could still sign the contributor agreement https://www.python.org/psf/contrib/contrib-form-jython (sorry for repeating myself)

jeff5:
Raised as jythontools/jython PR 28

PR #29, opened 2016-01-14
Fix netty ssl bug.

nickmbailey:


jimbaker:
Merged in https://hg.python.org/jython/rev/3983294381a0

PR #30, opened 2016-02-01
Fix resource closer synchronization.

nickmbailey:


Stewori:
This patch looks like it would revive [Issue 2280](http://bugs.jython.org/issue2280).
(It's not easy to cover this issue by tests, because in its nature as a synchronization issue it only occurs sporadically.)
Synchronization on PySystemStateCloser.class should be touched with care; note that FinalizeTrigger.runFinalizer also synchronizes on PySystemStateCloser.class.
To get an overview of the impact see [the patch for issue 2280](https://hg.python.org/jython/rev/f79dc13778f9).
Maybe we can find a way to fix issue 2457 without breaking the that patch...?

nickmbailey:
I think its possible this doesn't revive 2280. registerCloser and unregisterCloser are no longer synchronized since the set of closers is now a synchronized collection. So the finalizer thread at least won't be locking on PySystemStateCloser. Although I suppose the deadlock might just move to within the synchronized collection...

I think ultimately this code probably needs a bit of a larger refactoring though.

Stewori:
I think a safe fix would be to lock org.python.modules._io.Closer on PySystemStateCloser.class too, instead of Closer.class. I.e. remove synchronized from Closer.dismiss and Closer.call and instead insert inner synchronized(PySystemStateCloser.class)-blocks into these methods.
This should avoid at least this deadlock, but is indeed no solution for other users of the PySystemStateCloser API.

PySystemState.registerCloser and related API should document that Closer-implementers must not introduce their own locks due to risk of deadlocking. Instead they should lock on PySystemStateCloser.class. (Yes this looks like a GIL, but only scopes closing and finalization phase which might not be too bad.)

Your solution might work as well, however I currently cannot assess the whole impact of the synchronization changes you suggest (The deadlock might move as you point out).

And you are right - the finalization API was designed without the PySystemStateCloser-API in mind, which also caused 2280. It should be adjusted to incorporate this. I'd suggest to apply the safe solution for 2.7.1 and target a better solution for 2.7.2, unless @jimbaker spontaneously sees a better solution.

nickmbailey:
I wonder if the right solution isn't just to remove the synchronized attribute on Closer.dismiss(). If call() stops setting sys to null dismiss can safely be called numerous times. Closer.call() can either use some other instance variable for preventing repeated work or just check client.get() multiple times. I'm not sure how expensive it is.

Stewori:
Instead of checking sys == null we would need an atomic boolean in order to safely drop synchronized from dismiss. E.g.:

 AtomicBoolean dismissed = new AtomicBoolean();
 public synchronized void dismiss() {
if (dismissed.compareAndSet(false, true))
            sys.unregisterCloser(this);
    }
    public synchronized Void call() {
    	if (dismissed.compareAndSet(false, true)) {
             ...

nickmbailey:
Well unregisterCloser should be idempotent so I don't think we need any synchronization at all. It can be called safely no matter what.

Stewori:
That's probably true. My concern was that - consistently with the old implementation - dismiss and call should not both be sucessfully called. If dismiss is called it can expect that the closer is not unregistered during its process. However I agree this would be unlikely to become an issue.
Still the dismiss-boolean is needed, but than can be an ordinary boolean given that call continues to be synchronized. call does not potentionally call into PySystemStateCloser.class-locked API, does it? Maybe it would be safer to just lock it on PySystemStateCloser.class too.

Stewori:
Reading the code again, there is the comment

// This will prevent repeated work and dismiss() manipulating the list of closers

So the original author had some reason to avoid allowing a call to sys.unregisterCloser after processing of the closer via "call" had started. So I'd prefer to stick to the AtomicBoolean variant like printed above also given that the performance difference should be negligible.
@nickmbailey Could you please test whether that variant solves the issue?
If yes, I would merge it after a review from Jim.

PR #32, opened 2016-02-08
2461 - Respect getpeername returning tuple with 4 values

teeohhem:


Stewori:
Was merged some time ago as of https://hg.python.org/jython/rev/3881da5b10a8.
However, by now ssl.py has changed rather much; _createSSLEngine looks like it has been re-implemented completely. I guess there would have been an issue filed if the new implementation had re-introduced issue 2461, so I suppose

this PR can be closed.

PR #36, opened 2016-03-02
Add PosixModule.uname() compliant with posix *and* Windows

Stewori:
If there are no concerns I'd like to merge this uname implementation.
I tried to resemble the semantics of uname for windows to improve portability (any comments on this?) - with windows and posix support the vast majority of platforms should be covered.


Stewori:
Note that I did not use Java property os.arch, because it sometimes evaluates to x86_64 and sometimes to amd64, while uname -m provides more consistent output. This is more suitable in case some script performs a string comparison against it.

Stewori:
I reworked the approach a bit. Now machine-name (uname -m) is normalized to usual uname -m result string, also on Windows given the value of PROCESSOR_ARCHITECTURE. Also, uname -n is now emulated by InetAddress.getLocalHost().getHostName(). The former approach is kept as a fallback.
Except for uname -v the emulation should even work to some extend on systems that are neither POSIX (i.e. lack native uname command) nor Windows (does such system exist at all?)

Stewori:
Note that this PR is required to use [JyOpenGL](https://github.com/Stewori/JyOpenGL) with [JyNI](https://github.com/Stewori/JyNI.)

Stewori:
Committed as of aaec0b157cd3ff292e39fcaf14344e9a05d58abc.

PR #37, opened 2016-03-10
Wait should always joins the coupler threads

cwickl:
If a caller is using poll(), and the poll() sets returncode != None, then a subsequent call to wait() will never join the coupler threads.
In the case that the execution context is redirecting output to files (e.g. if the process is expected to generate large amounts of output), the coupler threads may not yet be complete.
At that point, the caller cannot be certain that the output buffers are flushed (the implication being that the caller owns the I/O streams, but has no way to tell when they are flushed - the process may no longer be running, but only wait() will currently join the couplers).

### DIFF
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -1426,13 +1426,13 @@ class Popen(object):
             attribute."""
             if self.returncode is None:
                 self.returncode = self._process.waitFor()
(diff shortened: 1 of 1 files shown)


Stewori:
Was merged some time ago as of https://hg.python.org/jython/rev/35fa19ca1859.
Can be closed.

PR #38, opened 2016-04-22
Adjust sysconfig.py to comply with a hack regarding platform detection in JyNI.

Stewori:
Adjust sysconfig.py to comply with a hack regarding platform detection in JyNI. Changes are minimal and don't touch ordinary Jython oparation at all. Only some checks are done more explicitly now, i.e. it confirms os.name == 'java' here and there, which is true for Jython anyway.


Stewori:
Committed (with minor further adjustment) as of 92c51dcd0aa9151f0c5a374474baccc8fe662c99.

PR #40, opened 2016-05-11
2500 - Don't search for more cacerts if one is passed in. Also be more careful when loading possible cacert files

teeohhem:


darjus:
Pushed in hg

PR #41, opened 2016-06-18
Allow __doc__memberName to be a String as well as a PyString.

shoaniki:
Allowing docstrings to be Java strings is is desirable for pure Java code that wants to be friendly to both Jython and Java developers, by making docstrings available for class members without introducing a hard dependency on the Jython runtime.

This would close the venerable [issue 608632](http://bugs.jython.org/issue608632).  There's an old patch there that allowed the use of any object type, which seems unnecessary.


Stewori:
I would like to accept this one. @shoaniki, did you sign the [contributor agreement](https://www.python.org/psf/contrib/contrib-form-jython/)? Also, tell me your name so you can be added to the contributors-list in ACKNOWLEDGEMENTS-file.

Stewori:
Given that the change is fairly small I decided to just merge it. However I modified it to accept any java.lang.CharSequence (e.g. String, Stringbuffer, StringBuilder) after I let PyBaseString and thus PyString and PyUnicode as well implement CharSequence.
Still, I would recommend to sign the contributor agreement anyway. Also, feel free to tell me your name then I will add you to ACKNOWLEDGEMENTS-file.

Merged as part of https://github.com/jythontools/jython/commit/2a68fb90591556d0abafe211a9cc2067a8a74585
Can be closed.

PR #57, opened 2017-02-23
Fix #2534 Make os.getlogin() throw OSError if OS call fails

jamesmudd:
This makes the Jython behaviour consistent with CPython implementation
allowing calling code to handle the OSError portably across both
implementations.

The message informs of the user of the failure and suggests using
the environment variable instead, as suggested by the Python
documentation https://docs.python.org/2/library/os.html#os.getlogin


Stewori:
Merged with slight modification as of https://hg.python.org/jython/rev/07724b0547e0
- made the message a bit more formal
- Jython styleguide requires if always written as {} block

(not visible on github due to http://bugs.jython.org/issue2556)

Can be closed.

PR #58, opened 2017-02-28
#2558 Allow PyString to be coerced to an Enum when calling a method

jamesmudd:


jamesmudd:
This should probably have a test with it. I'm not sure where a test for this should be, could someone point me to other automatic type conversion tests?

Stewori:
Maybe Lib/test/test_str_jy.py is a good place to add a test for this..?
If your test incorporates Java-code that should go somewhere into tests/java AFAIK.
If you need a Java unittest (using junit) that should go into tests/java/javatests AFAIK.

jamesmudd:
Replaced by  https://github.com/jythontools/jython/pull/62

PR #50, opened 2016-11-12
Fix calling Java vararg methods with no arguments fails

jamesmudd:
Also add additional testing for vararg methods to improve test coverage
of boxing varargs.

PR #59, opened 2017-02-28
Update NEWS

jamesmudd:
Add https://github.com/jythontools/jython/pull/50 to news


Stewori:
Merged as of https://hg.python.org/jython/rev/4b96832f0b76.
Can be closed.

PR #61, opened 2017-03-01
Fix #2559 Increase default ThreadStackSize to -Xss2560m

jamesmudd:
This fixes a stack overflow seem in test_marshal this is significantly
larger than needed to allow the test to pass -Xss3m was needed so this
should be sufficient in all cases.


jamesmudd:
I have updated this to use a Xss of 2560k this is a little smaller but still sufficient to pass the test (At least where I have tried it). I have also rebuilt the Windows exe launcher with the increased Xss and tested this on Windows 10.

I think it would be a good idea for a few more people to try running this to ensure the launcher is robust.

darjus:
Don't have access to a Windows machine, maybe someone else can double-check. The change looks reasonable.

jeff5:
I'll take a look (Windows 10) but don't let that stop anyone else.

jamesmudd:
Have just updated this again sorry I missed the test_jython_launcher which explicitly checks the Xss value

jeff5:
LGTM on Windows 10 with Java 1.7.0_60 and Java 1.8.0_121. Java 7 runs clean (with known hangs excluded) and failures on Java 8 are pre-existing unrelated issues. Thanks for doing this. 👍 

Linux ok?

jeff5:
I'll commit this squashed, via hg, with James' title and identity.

PR #62, opened 2017-03-01
#2558 Allow PyString to be coerced to an Enum when calling a method

jamesmudd:
Also adds tests for this behaviour


jamesmudd:
This needs a bit more thought

PR #64, opened 2017-03-13
#2568 Add an epsilon to allow for timing inaccuracy

jamesmudd:
This is the same approach already used in this test (line 300). It is
required due to the timing accuracy of the JVM Thread.sleep
implementation


Stewori:
Merged as of https://hg.python.org/jython/rev/f06cf81336e0.

PR #65, opened 2017-03-15
# 2318 Fix test_zipimport_jy

jamesmudd:
This test was always failing on Windows, and would fail if run
individually on unix as it relied on the cwd to find a zip file. The
actual file wasn't important to the test, and is just used to create a
instance of the required class. Additionally the zip file used wasn't
included in the built /dist dir. The test now uses the same zip file as
used in other parts of this test.


Stewori:
Merged as of https://hg.python.org/jython/rev/3898500ffc7e.

PR #66, opened 2017-03-15
#2554 Fix test_classpathimporter fails on Windows

jamesmudd:
The issue here was the test relied on the cwd which is not set
consistently when running the tests in different ways.


jamesmudd:
The name of the function test_support.DirsOnSysPath is a little misleading in this case its used to add jar files to the path. This change doesn't alter that just though maybe it should be renamed.

Stewori:
Merged as of https://hg.python.org/jython/rev/e725a0b2d939.

PR #67, opened 2017-03-20
#2399 Fix test_sort failure on Java 8

jamesmudd:
This approach caches the exception raised in the comparator and then
allows the sort to continue. At the end it checks if any exceptions were
raised and re-throws them.

Also adds missing @Override in edited inner classes.


jeff5:
This looks good to me. And it makes a second Travis build go green. I'll merge via the repo at python.org, to close when that's done.

jeff5:
Now considered fixed as of https://hg.python.org/jython/rev/c677bb7414b1

PR #69, opened 2017-04-04
Correct white space formatting mistake

jamesmudd:
Noticed I had left a tab here after a earlier change


Stewori:
Merged as of https://hg.python.org/jython/rev/d8a6f64dd104

PR #70, opened 2017-04-04
Fix #2313 test_jython_initializer fails on Windows

jamesmudd:
This moves SyspathAppendingInitializer and the service definition for
org.python.core.JythonInitializer into a jar which can be more easily
bundled into the tests. The test is then modified to use the jar instead
of using a class file outside /dist which allows the path to resolve on
Windows.

The jar was built using Java 7 (1.7.0_80)

This also removes the need for a custom compile entry in the build
script.


jamesmudd:
@jeff5 Thanks for looking at this pull request. The source is still included in the jar. I don't think this one can be solved just by removing the unix paths, as the required files are not copied into the dist folder during the build. I could be wrong though maybe this isn't a issue? Would be interested to hear other opinions.

jeff5:
Ah, I understand now 🤦‍ :  it fails after installation, but not in regression tests, because it looks for `dist`. Ok, this _is_ better.

I've seen comments elsewhere that suggest we want to get rid of jars from the test tree, i.e. generate them in the test, but we haven't, and this is an improvement on having a failing test. I'll leave a couple days in case anyone else has a comment.

jeff5:
I found the place where a preference is expressed agains JAR files in the tree. It's here: https://hg.python.org/jython/file/tip/Lib/test/test_classpathimporter.py#l21. This is @fwierzbicki writing. Respectfully however, we've let it pass there for 8 years, so why not here?

Thanks James. This PR now in the hg repo at: https://hg.python.org/jython/rev/0a00982f6ea5 .

PR #75, opened 2017-05-04
Fix #2585 test_ssl test failure

jamesmudd:
This adds a work around for the Netty exception
AnnotatedNoRouteToHostException which extends NoRouteToHostException
which is explicitly handled by the exception mapping.


Stewori:
This looks good to me and I can confirm it fixes test_ssl failure on linux. I am going to accept it...

Stewori:
Merged as of https://hg.python.org/jython/rev/7ad8eedaa14a together with #77.

PR #76, opened 2017-05-04
Add a Java 8 linux build target on Travis

jamesmudd:


Stewori:
Maybe we should also add openjdk8-entry? Or does something speak against it?

jamesmudd:
Yes openjdk should be included as well. I will add it.

jamesmudd:
So the issue is that openjdk8 is not available on the Travis Precise environment we are currently using see https://github.com/travis-ci/travis-ci/issues/6483

We could change to the Trusty environment (but that is in beta) https://docs.travis-ci.com/user/trusty-ci-environment/ which would then allow openjdk8 to be used. Or we could ignore openjdk8 for the moment and just add oracle java 8. I think we definitely should have a Java 8 build on linux I suspect more people are building and running Jython on Java 8 now than on Java 7.

Note: The MacOS build which uses Travis default JDK is using Java 8.

Stewori:
Alright, let's keep OpenJDK 8 out of this for now; I was just curious why it wasn't included. Your explanation sound convincing.
The remaining failures I observe on Linux http://bugs.jython.org/issue2587 and http://bugs.jython.org/issue2586 are mainly file-system specific (NTFS). 2586 has also some issue with date, maybe it's time-zone specific. So, I guess we could give it a try on travis.

jamesmudd:
I have changed this to only have Oracle Java 8. I'm a little confused by the build log output https://travis-ci.org/jythontools/jython/jobs/229876445 it seems like something at the start of the build is still using java 7 (see line 158). But the tests are run under java 8 (see line 526). I'm not actually sure if the build is done using java 7 or 8.

darjus:
@jamesmudd Your changes look good. The initial java version is before we switch the JDK ( https://travis-ci.org/jythontools/jython/jobs/229876445#L526 )

darjus:
Merged in hg.python.org

PR #77, opened 2017-05-04
2585 2

jamesmudd:


Stewori:
Merged as of https://hg.python.org/jython/rev/7ad8eedaa14a together with #75.

PR #81, opened 2017-07-05
Minor improvement to PyType to use computeIfAbsent

jamesmudd:


jamesmudd:
Related to http://bugs.jython.org/issue2487 so someone who knows that well might want to check this

jamesmudd:
Abandon this would be better, but would require Java 8. Is there a strong reason not to move Jython to Java 8 now?

PR #82, opened 2017-07-09
Revert "Isolate each JSR ScriptEngine with separate PySystemState. Fi…

jamesmudd:
…xes #2154"

This reverts commit c7a3ebac42ba2efb6f235c0ec4a72c048d925fb3.


jamesmudd:
Discussed on http://bugs.jython.org/issue2604 this is not the correct solution

PR #85, opened 2017-07-24
Make fields used for synchronisation final to protect against NPE

jamesmudd:
Noticed this while running jython with the NPE breakpoint on. Occasionally my app will jump to the jython gc class when a NPE is thrown. It actually appears this is expected and is used for flow control, generally not the best approach. However in this commit I remove the flow control via NPEs and make the fields final which is better as they are synchronised on. The chnage seems to pass the regrtests ok, but maybe I have missed a important reason why it was done like this so someone who knows better should definitely review this.


Stewori:
While the changes look good, after touching gc-stuff one should always run `test_gc.py`, `test_gc_jy.py` and `test_JyNI_gc.py` (from JyNI) 200-300 times in a loop, possibly on slow and fast systems respectively.

On Linux I use kind of the following sh-script for this:

```
#!/bin/bash
x=1
while :
do
  bin/jython Lib/test/test_gc.py
  bin/jython Lib/test/test_gc_jy.py
  echo "tested $x times"
  x=$(( $x + 1 ))
  echo "======================================="
  echo ""
done
```

Note that you will have to hold ctrl-C for a second or so in order to exit the script.

If I find time I will also run the loop on my machines for a while, but as I said - the changes look good.
Originally I wrote the lazy map-creation, because the maps are only needed for a rarely used feature and I didn't want to increase the memory footprint of GC for common use cases. I suppose a check like `if (xy == null) {...}` would have been better than the exception based check. However, your changes simplify it much and the additional memory for these maps is probably negligible anyway.
If the loop-tests go well I will accept it.
Note that even the current implementation is not perfect, see [issue 2519](http://bugs.jython.org/issue2519). So if rare failures should occur it might be still okay, but let's see...

jeff5:
+1 for the improved construction using final and consequent simplification to the code.

Out of curiosity, I looked to see what space the unused containers cost us. HashSet : 5 ints, 2 refs; ArrayList: 1 int, 1 ref. Cleverly, in spite of what they say about their default initial capacity, neither container actually allocates that storage until the first element arrives. Each just has a reference to the same, zero length, array. Nice touch.

As these objects are singletons, I think the code required to be lazy is bigger than their footprint, and no-one will notice the contribution to start-up time.

"sync**h**ronisation" or "sync**h**ronization" _(Am.)_ by the way.

jamesmudd:
Thanks for taking a look at this. Hopefully I will get some time to look at it again next week. It does provide quite a bit of simplification. I will run the suggested tests and report back. @jeff5 Thanks for catching the typo, that should be fixed before this is merged!

jamesmudd:
I have been running the tests in a loop on my machine for a while I have completed 1250 iterations of `test_gc.py` and `test_gc_jy.py` without any failures so far. I will get it running on a raspberry pi for "slow" system testing (and it's single core, not sure if that matters). Hopefully I will get the JyNI test to run as well. I have fixed the synchronisation typo.

jamesmudd:
I have completed 500 iterations of the tests `test_gc.py` and `test_gc_jy.py` on a raspberry pi without any failures or errors.

Stewori:
Brilliant. Thanks a lot. I think I can merge it as soon as I find time. I can maybe do a JyNI check too.

Stewori:
Merged as of https://github.com/jythontools/jython/commit/9eaee1f576250dbb6b77cfee8a40fd5db9478d24. Closing...

PR #89, opened 2017-09-07
#2620 Fix "Error loading Python DLL" on Windows 7

jamesmudd:
The problem is caused by the python27.dll file being external to the
.exe See
https://stackoverflow.com/questions/19225132/pyinstaller-not-working-on-simple-helloworld-program

The fix here is to use PyInstaller with the "--onefile" or "-F" option.

This is also tidier as it results in having a combined executable.

Built with PyInstaller 3.2.1 on Python 2.7.13 on Windows-10-10.0.15063


jeff5:
@jamesmudd I agree. This came up elsewhere but you found a better how-to than I did. I think you may have done this right originally and I broke it when I rebuilt the launcher. The root of the problem, I suggest, is not having a dev guide (or ant target) for the launcher.

jamesmudd:
@jeff5 I agree the issue is there should be an ant target to generate this launcher. I will have a go at setting that up. Additionally I don't think there are any actual tests which try to use this launcher but that's a harder problem as we would need windows testing to work, preferably on all versions of Windows that are supported (I'm not even sure which Jython claims to support).

jeff5:
Thanks James.

`sys.executable` points to it, so test_subprocess etc. will exercise it. Almost by definition, we do not catch the problem fixed here.

jeff5:
Now at https://hg.python.org/jython/rev/d638b2c5ef28.

I think we could live without an ``ant`` target (and anyway, PyInstaller wll not be available for all). I'll add a note in the devguide and jython.py, which will make it necessary I go through the process myself.

jamesmudd:
@jeff5 Thanks for merging this. I have taken a look at the devguide and it looks good to me. I did have a look at the ant target, but it would be difficult as you mention it would need to assume you already have PyInstaller present and probably guess the path to it assuming default install directories.

PR #91, opened 2017-10-02
Fix #2403 VerifyError when implementing interface default methods

jamesmudd:
There are 2 commits the first adds a failing test for this then the second resolves it.
I don't think this should be merged as is. In order for the test to compile the source level was increased to 1.8 when default methods were added, as we still support 1.7 this is not acceptable.

However the second commit fixes the issue by increasing the Opcode to v1.6 this appears to resolve the issue with the test and passes the regrtests. I don't understand why. Increasing it higher 1.7 or 1.8 cause failures.

The fix commit could be merged without the test which would improve usability with Java 8


shoaniki:
The reason this works is that the change switches the generated class file version from 49 to 50, which happens to be the point at which the JVM starts using the up-to-date type-checking bytecode verifier instead of the old type-inferring one.  Apparently the old verifier does not accept the code Jython emits when implementing a default method (specifically, the instruction that attempts to call the default implementation if the Python class did not override it), while the new verifier recognises that it is legal.

Having this fix merged would be very useful, if doing so without the test is an option.  If the risk associated with a change in bytecode verifier is unpalatable, perhaps the opcode version could be made configurable?

jamesmudd:
I have just pushed an update to this pull request that doesn't include the test and therefore the Java 8 requirement. I have been playing with this a bit more and I would suggest it should be merged, it fully fixes this problem in every case I can think of. As discussed above, this works as it changes the bytecode verifier used to the current one. I think the risk associated with this is small I can't think of any case where this would cause problem. However would be good if someone else could have a look at this to check.

jeff5:
Thanks @jamesmudd  and @shoaniki. I'll try this out.

jeff5:
Now in at https://hg.python.org/jython/rev/f842e0a9d903, and https://github.com/jythontools/jython/commit/a083743246bcf64f05e4e1b06c109d71cf309d03, thanks again.

PR #93, opened 2017-11-02
Make the calendar used by the time module be Gregorian all the way to match CPython's implementation

tombech:
Before the patch:
>>> time.gmtime(-12212553600)
time.struct_time(tm_year=1583, tm_mon=1, tm_mday=1, tm_hour=0, tm_min=0, tm_sec=0, tm_wday=5, tm_yday=1, tm_isdst=0)
>>> time.gmtime(-12212553601)
time.struct_time(tm_year=1582, tm_mon=12, tm_mday=31, tm_hour=23, tm_min=59, tm_sec=59, tm_wday=4, tm_yday=355, tm_isdst=0)
>>>

After the patch it matches Python 2.7, Python 3.6 and PyPy:
>>> time.gmtime(-12212553600)
time.struct_time(tm_year=1583, tm_mon=1, tm_mday=1, tm_hour=0, tm_min=0, tm_sec=0, tm_wday=5, tm_yday=1, tm_isdst=0)
>>> time.gmtime(-12212553601)
time.struct_time(tm_year=1582, tm_mon=12, tm_mday=31, tm_hour=23, tm_min=59, tm_sec=59, tm_wday=4, tm_yday=365, tm_isdst=0)
>>>

### DIFF
--- a/src/org/python/modules/time/Time.java
+++ b/src/org/python/modules/time/Time.java
@@ -226,12 +226,15 @@ public class Time implements ClassDictInit
     }
 
     private static GregorianCalendar _tupletocal(PyTuple tup) {
-        return new GregorianCalendar(item(tup, 0),
-                                     item(tup, 1),
-                                     item(tup, 2),
-                                     item(tup, 3),
-                                     item(tup, 4),
-                                     item(tup, 5));
+        GregorianCalendar gc = new GregorianCalendar(item(tup, 0),
+                                                     item(tup, 1),
+                                                     item(tup, 2),
+                                                     item(tup, 3),
+                                                     item(tup, 4),
+                                                     item(tup, 5));
+                                                     
+        gc.setGregorianChange(new Date(Long.MIN_VALUE));
+        return gc;
     }
 
     public static double mktime(PyTuple tup) {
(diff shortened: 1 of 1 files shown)


jamesmudd:
This seems like a good fix to me. The relevant docs explaining the existing Python/Java discrepancy are here: [Python](https://docs.python.org/2/library/datetime.html#date-objects) and [Java](https://docs.oracle.com/javase/7/docs/api/java/util/GregorianCalendar.html#setGregorianChange(java.util.Date))

Would be nice to add a test for this. I would suggest [test_datetime_jy.py](https://github.com/jythontools/jython/blob/master/Lib/test/test_datetime_jy.py) in [TestCalendar](https://github.com/jythontools/jython/blob/bb2431d434bfcd85d72132fcacca1471fa00ae4d/Lib/test/test_datetime_jy.py#L13)

@jeff5 I think this should be merged could you have a look.

jeff5:
I didn't immediately find that CPython supports the conversion @tombech shows, but this just CPython on Windows. Jython is already not restricted in range, so let's get proleptic.

We ought to test it too. It is surprising how little time.gmtime is actually tested. test_time.py seems to be the place for a test, which we have already customised a little to _tolerate_ Jython's increased range:
https://github.com/jythontools/jython/blob/b91048cd514a0aafa357ccceb8939e44794d29db/Lib/test/test_time.py#L48-L52

The thing that maybe encapsultes this PR is where a one second difference takes an extra 10 days:
```
>>> t = -12219292800
>>> time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(t))
'1582-10-15 00:00:00'
>>> time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(t-1))
'1582-10-04 23:59:59'
```
This is correct by some calendars (and not others), but Python wants to make that your problem, while Java assumes we're early adopters. :-) I hope I'm correct to ignore time zones here.

jeff5:
Now in trunk at https://hg.python.org/jython/rev/db9b69e4e8ef and a test in the changeset following. Thanks again to @tombech .

PR #97, opened 2017-12-08
#2650 Initialise detailMessage for PyException

jamesmudd:


jeff5:
I think this approach to exceptions, deferring the creation of a message, is deliberate.

Exceptions in Python are a normal means of control transfer, whereas in Java they're expicitly not. (I could find the refs, I think.) Once we come to print the thing out, of course, then we know it's not just a transfer of control, but I think we're putting it off until then.

Your use case is good, however. My instinct is that the exception ought not to escape an embedded interpreter without a proper message.

We get to do that for the JSR223 engine here, by wrapping:
https://github.com/jythontools/jython/blob/6804dfbadf78462a66ff3d4113f7ce5bfd835a17/src/org/python/jsr223/PyScriptEngine.java#L194

But `PythonInterpreter` gives you the actual `PyException`, being a thin veneer over `PySytemState` and the built-in `eval`. 

Would it work to define ``PyException.getMessage()`` so it yields the text you hope for?

jamesmudd:
@jeff5 Thanks for the comments. Yes overriding `PyException.getMessage()` would also work, and would probably be preferable in this case for the performance reason mentioned.

I implemented it in a Java way i.e. exceptions shouldn't be used for flow control, therefore if you are creating one you should make it contain all the information it needs and ideally make it immutable.

My only question would be is making the detail message noticeable when you consider that your about to create the Java stack trace anyway? I would have guessed that took much longer? But for now I will change this to lazily build the message.

jeff5:
Yes, it's more natural for Java, but the (C)Python way is influenced by good old `errno`. We simply keep all the broken pieces in the thread state, and I think there's no real exception object until it's caught. The `PyException` isn't a `PyObject`.

The Java stack trace might be no more than a pointer to the last frame at this stage. The Python one is a linked list of names, and no-one does any further text formatting until asked:
https://github.com/jythontools/jython/blob/6804dfbadf78462a66ff3d4113f7ce5bfd835a17/src/org/python/core/PyTraceback.java#L25-L30
The most wasteful part may be the way we seem to  re-throw the (Java) exception to work our way our of each (Python) frame, but I guess in that case you probably are heading for a stack trace not a control transfer.

jamesmudd:
I have updated this PR to override `getMessage` instead of initializing the `Throwable` `detailMessage` field.

jeff5:
Thanks James! Now committed at: https://hg.python.org/jython/rev/d74f8c2cd56f and https://github.com/jythontools/jython/commit/d22144441f9a0ba76c9d87ee24d4a537a15423f8

PR #45, opened 2016-08-01
Upgrade netty master

nickmbailey:

PR #98, opened 2018-01-26
Remove limitation

pwillemse:
When trying to send snmp messages I discovered that jython was hanging when sending messages > 64k. In jython 2.7.0 it worked fine. In _socket.py a limitation on the message size was found. Why just truncating a message and not handling the rest of the message? After removal it worked fine.


Stewori:
For me this looks like ``self.channel.bytesBeforeUnwritable()`` returns a too restrictive value. I suspect that ``bytesBeforeUnwritable()`` does not exist for nothing, maybe a better solution would be to fix it. However, I'm no expert in this area and must admit the I actually don't know what ``bytesBeforeUnwritable()`` is about and what consequences it can have to ignore it.

pwillemse:
Why was this code not present in jython 2.7.0 and added in 2.7.1? Was this only because some one was worried about the size of the packets?
One thing for sure: SNMP is broken when sending packets > 64k.

jeff5:
The change that added this (https://github.com/jythontools/jython/commit/b348d77b03f4e2f518b7b6b714e93104103e3ebe and #45) is offered as a fix for http://bugs.jython.org/issue2508, and the present PR essentially reverts the code change in that fix, re-opening the bug. This may be an improvement, but it isn't going to provide correct behaviour.

It was pointed out later that the code change in #45 has introduced another problem http://bugs.jython.org/issue2618, which I guess is what the OP is struggling with. @Stewori : I'd say `bytesBeforeUnwritable` exists for a reason, but perhaps not the one we're using it for here.

@jimbaker : Is it possible we are providing non-blocking semantics in circumstances where an application really needs the library to block?

jeff5:

PR #102, opened 2018-05-17
#2672 Fix incorrect formatting of Integer.MIN_VALUE

jamesmudd:


Stewori:
Merged as of https://hg.python.org/jython/rev/123a2956db3f.
Thanks again for this work.
Minor edits:
Use exact title from BJO in News file.
Added issue link to each new test, not only to the first one (that was supposed to denote several tests below but might become unclear once new tests would be appended, so one comment for each test is clearer).

PR #111, opened 2018-10-12
zlib decompress fixes

devendor:
This allows proper processing of gzip trailer rather than leaving it in
the queue for resubmission.  This allows msgpack and pip-18 to work with
a few other workarounds submitted to upstream msgpack.

Version with workarounds is https://github.com/devendor/pip and
https://github.com/devendor/msgpack-python.  The latter has been
accepted in upstream master and will eventually replace the
_vendor/msgpack shadowed in pip.

### DIFF
--- a/Lib/test/test_zlib_jy.py
+++ b/Lib/test/test_zlib_jy.py
@@ -51,7 +51,7 @@ class ArrayTestCase(unittest.TestCase):
         self.assertRaisesRegexp(zlib.error, 'Error -3 while decompressing data: incorrect length check',
                                 dco.decompress,c)
 
-    def test_decompress_badlen(self):
+    def test_decompress_badcrc(self):
         """Manipulating last crc bytes to create a crc check exception.
         RFC-1952:
            0   1   2   3   4   5   6   7
--- a/Lib/test/test_zlib_jy.py
+++ b/Lib/test/test_zlib_jy.py
@@ -28,6 +28,44 @@ class ArrayTestCase(unittest.TestCase):
         compressed = array('c', compress('jython'))
         self.assertEqual('jython', decompress(compressed))
 
+    def test_decompress_gzip(self):
+        co = zlib.compressobj(wbits=31)  # window 15 with gzip wrapper.
+        c = co.compress("Jenny: 867-5309")
+        c += co.flush()
+        dco = zlib.decompressobj(wbits=31)
+        d = dco.decompress(c)
+        self.assertEqual(b'',dco.unused_data, msg="dco.unused_data not empty after decompress.")
+        self.assertEqual(b'',dco.unconsumed_tail, msg="dco.unconsumed_tail not empty after decompress.")
+        self.assertEqual("Jenny: 867-5309",d)
+
+    def test_decompress_badlen(self):
+        """Manipulating last two bytes to create invalid initial size check.
+        RFC-1952:
+           0   1   2   3   4   5   6   7
+         +---+---+---+---+---+---+---+---+
+         |     CRC32     |     ISIZE     |
+         +---+---+---+---+---+---+---+---+turn:
+        """
+        c=b'\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\x03\x0bJ\xacT(O,V\xc8H-J\x05\x00\xc2\xb0\x1e\xe5\x0d\x00\x00\x00'
+        dco = zlib.decompressobj(wbits=31)
+        self.assertRaisesRegexp(zlib.error, 'Error -3 while decompressing data: incorrect length check',
+                                dco.decompress,c)
+
+    def test_decompress_badlen(self):
+        """Manipulating last crc bytes to create a crc check exception.
+        RFC-1952:
+           0   1   2   3   4   5   6   7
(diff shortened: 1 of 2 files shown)


Lib/zlib.py
```diff
@@ -224,10 +228,20 @@ def decompress(self, string, max_length=0):
 
         self.inflater.setInput(string)
         inflated = _get_inflate_data(self.inflater, max_length)
+        self._crc32.update(inflated)
 
         r = self.inflater.getRemaining()
         if r:
-            if max_length and not self.inflater.finished():
+            if self.gzip and self.inflater.finished() and r == 8:
```
devendor: If we don't also strip off the trailer and leave it in unused_data, programs that expect the stream to be emptied keep resubmitting unused  data which just throws a warning about not enough bytes to decode until a recursion limit is exceeded.  This does one better by validating the content size and crc.

devendor (commented):
Just some inline comments to ease review of the changes.

Stewori:
This looks good to me, however a definite proof of concept in form of a test would be perfect.
If a test could be provided demonstrating the issue fixed, i.e. failing before the fix and passing afterwards I think this could be merged.

devendor:
@Stewori This one is much simpler than the others and I've added three tests for the change so it should be ready.

Lib/zlib.py
```diff
@@ -165,7 +164,7 @@ def flush(self, mode=Z_FINISH):
         if mode == Z_FINISH:
             if self._gzip:
                 last += self.GZIP_TRAILER_FORMAT.pack(
-                    self._crc32.getValue(), self._size % sys.maxint)
+                    self._crc32.getValue(), self._size % 2**32)
```
jeff5: You're right. Probably the author intended  `self._size & sys.maxint`, equivalent to the modulus operation. `&` is usually a bit quicker than `%`, but it's not an issue here.

Also, I would not like to guarantee that we're not computing 2**32 every time.

devendor:
I think I have that worked out at this point. It's under Raymond Ferguson and Devendortech LLC.  The latter is just 1 person LLC in Wisconsin I use professionally.

Lib/zlib.py
jeff5: I tweaked your patch a little so that here and elsewhere `% 2**32` becomes `& _MASK32`. Even if it isn't significant for performance, the idea we're computing 2**32 each time and doing long-division just sticks a bit in the throat.

Lib/test/test_zlib_jy.py
```diff
@@ -28,6 +28,44 @@ def _test_array(self, compress, decompress):
...
+        c = co.compress("Jenny: 867-5309")
+        c += co.flush()
+        dco = zlib.decompressobj(wbits=31)
+        d = dco.decompress(c)
+        self.assertEqual(b'',dco.unused_data, msg="dco.unused_data not empty after decompress.")
+        self.assertEqual(b'',dco.unconsumed_tail, msg="dco.unconsumed_tail not empty after decompress.")
+        self.assertEqual("Jenny: 867-5309",d)
+
+    def test_decompress_badlen(self):
+        """Manipulating last two bytes to create invalid initial size check.
```
jeff5: Documentation strings on test functions pop out on the console when that test runs, and long ones come out as a real mess. I converted these to comments.

jeff5 (commented):
Notes on tweaks I made. (Not a change request.)

jeff5:
I've committed this via mercurial (as we do) here: https://hg.python.org/jython/rev/c25c2edd4b16 (with a few tweaks). As a result, we have some merge conflicts as far as git is concerned. However, these are not to be resolved.

@devendor : thanks for doing this this. I believe you could delete your `zlib_fix branch` whenever you like.

devendor:
@jeff5 @Stewori thank you for the cleanup on this.  Sorry for the stall in progress.  I need to balance the effort with paying the bills.

jeff5:
Thanks @Stewori . I had a nagging feeling I'd missed something. 

We do not have a standard way to refer to issues raised and fixed via GitHub. GH-111 is succinct, but I think that this repo is actually just a temporary home, and that after a move, numbering will begin again at 1. Just the number works for bugs.jython.org because we all know the context. Ok for now, I guess.

Stewori:
I used the PR### notation several times before to incorporate commits into NEWS that were not tracked in bjo.
E.g. ``[ PR28 ] zipimporter supports multi-byte characters in pathnames``
So far nobody complained. If we should want to change this notation, it should be changed consistently. But honestly -- let's keep things simple and just keep it this way. We can add a clarifying comment on top that PR### numbers refer to https://github.com/jythontools/jython.

> but I think that this repo is actually just a temporary home

Seriously; there should not be a future home for Jython 2.7 on Github. Let's start things over for Jython 3 only (another story).

Stewori:
Only thing is, we might want to move Jython 2.7 from jythontools to jython organization. I think jythontools was only a temporary organization name because the name jython was occupied by some third party. However it should be possible to move it and keep existing bugtracker, PRs, commits valid: https://help.github.com/articles/transferring-a-repository-owned-by-your-organization/#transferring-to-another-organization

jeff5:
Raised as jythontools/jython pull 111

PR #112, opened 2018-10-12
A series of patches to allow jython to work with msgpack without

devendor:
workarounds in upstream msgpack. This also allows pip-18.1
compatitiblity by way of dependency when combined with the Lib/zlib.py
GZIP_TRAILER fix in decompress.

* src/org/python/core/Py2kBuffer.java

This is fairly cosmetic in that it moves __str__() functionality to
toString() and fixes the intended __repr__() customization that matches
cython and was broken in the 11 year old __repr__() derived object
recursion overflow workaround.

* src/org/python/core/PyMemoryView.java

This actually restricts and removes problematic functionality to make it
behave like cython which rejects attempts to take a memoryview of a
unicode object with a TypeError exception.  This is important for
pip/msgpack compatibility due to the fact that TypeError triggers use of
buffer to wrap the bytes in utf-16-be encoding rather than attempting to
strip it off entirely and potentially throw a unicode decode error.

* src/org/python/core/PyByteArray.java

This allows bytearray objects to __iadd__ PyMemoryView objects.
b=bytearray('asdf')
b+=memoryview('jkl;')

This is used by msgpack and supported by cython.

This also switches toString to create a string appropriate for java and
moves __repr__ functionality into bytearray___repr for cleaner repr
support and using the (String) value in struct.unpack methods.

* src/org/python/modules/struct.java

This set of changes allows struct to unpack from Py2kBuffer aka buffer
which is support by cython 2.7.x and used as a workaround for cython <=
2.7.6 which does not support unpacking from a bytearray.

It also adds support for unpacking from a bytearray to provide
compatibility without the workaround in msgpack.

### DIFF
--- a/src/org/python/core/PyByteArray.java
+++ b/src/org/python/core/PyByteArray.java
@@ -1325,8 +1325,10 @@ public class PyByteArray extends BaseBytes implements BufferProtocol {
         } else if (oType == PyString.TYPE) {
             // Will fail if somehow not 8-bit clean
             setslice(size, size, 1, (PyString)o);
+        } else if (o instanceof PyMemoryView){
+            setslice(size, size, 1, (BufferProtocol)o);
         } else {
-            // Unsuitable type
+	    // Unsuitable type
             throw ConcatenationTypeError(oType, TYPE);
         }
         return this;
@@ -1996,12 +1998,17 @@ public class PyByteArray extends BaseBytes implements BufferProtocol {
      */
     @Override
     public String toString() {
-        return bytearray_repr();
+        return this.asString();
+    }
+
+    @Override
+    public PyString __repr__(){
+       return bytearray___repr__();
     }
 
     @ExposedMethod(names = {"__repr__"}, doc = BuiltinDocs.bytearray___repr___doc)
-    final synchronized String bytearray_repr() {
-        return basebytes_repr("bytearray(b", ")");
+    final synchronized PyString bytearray___repr__() {
+        return new PyString(basebytes_repr("bytearray(b", ")"));
     }
 
     /**
--- a/src/org/python/core/PyMemoryView.java
+++ b/src/org/python/core/PyMemoryView.java
@@ -67,7 +67,8 @@ public class PyMemoryView extends PySequence implements BufferProtocol, Traverse
         ArgParser ap = new ArgParser("memoryview", args, keywords, "object");
         PyObject obj = ap.getPyObject(0);
 
-        if (obj instanceof BufferProtocol) {
+        // memoryview in 2.7 throws typeerror on unicode rather than assuming encoding like buffer().
+        if (obj instanceof PyString && !(obj instanceof PyUnicode)) {
             return new PyMemoryView((BufferProtocol)obj);
         } else {
             throw Py.TypeError("cannot make memory view because object does not have "
--- a/src/org/python/modules/struct.java
+++ b/src/org/python/modules/struct.java
@@ -7,7 +7,6 @@
  */
 
 package org.python.modules;
-
 import org.python.core.Py;
 import org.python.core.PyException;
 import org.python.core.PyFloat;
@@ -21,6 +20,8 @@ import org.python.core.PyTuple;
 import java.math.BigInteger;
 import org.python.core.ClassDictInit;
 import org.python.core.PyArray;
+import org.python.core.PyByteArray;
+import org.python.core.Py2kBuffer;
 
 /**
  * This module performs conversions between Python values and C
@@ -1077,7 +1078,7 @@ public class struct implements ClassDictInit {
      * The string must contain exactly the amount of data required by
      * the format (i.e. len(string) must equal calcsize(fmt)).
      */
-   
+
     public static PyTuple unpack(String format, String string) {
         FormatDef[] f = whichtable(format);
         int size = calcsize(format, f);
@@ -1096,11 +1097,36 @@ public class struct implements ClassDictInit {
             throw StructError("unpack str size does not match format");
          return unpack(f, size, format, new ByteStream(string));
     }
-    
-    public static PyTuple unpack_from(String format, String string) {
-        return unpack_from(format, string, 0);   
+
+    public static PyTuple unpack(String format, Py2kBuffer buffer){
+        return unpack(format, buffer.toString());
     }
-        
+
+    public static PyTuple unpack(String format, PyByteArray bytearray) {
+        /* bytearray is added in 2.7.7 */
+        return unpack(format, bytearray.toString());
+    }
+
+    public static PyTuple unpack_from(String format, Py2kBuffer buffer) {
+        return unpack_from(format, buffer.toString(), 0);
+    }
+
+    public static PyTuple unpack_from(String format, PyByteArray bytearray) {
+        return unpack_from(format, bytearray.toString(), 0);
+    }
+
+    public static PyTuple unpack_from(String format, Py2kBuffer buffer, int offset) {
+        return unpack_from(format, buffer.toString(), offset);
+    }
+
+    public static PyTuple unpack_from(String format, PyByteArray bytearray, int offset) {
+        return unpack_from(format, bytearray.toString(), offset);
+    }
+
+    public static PyTuple unpack_from(String format, String string) {
+        return unpack_from(format, string, 0);
+    }
+
     public static PyTuple unpack_from(String format, String string, int offset) {
         FormatDef[] f = whichtable(format);
         int size = calcsize(format, f);
@@ -1109,7 +1135,7 @@ public class struct implements ClassDictInit {
             throw StructError("unpack_from str size does not match format");
         return unpack(f, size, format, new ByteStream(string, offset));
     }
-    
+
     static PyTuple unpack(FormatDef[] f, int size, String format, ByteStream str) {
         PyList res = new PyList();
         int flen = format.length();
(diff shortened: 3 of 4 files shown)


Stewori:
@devendor This looks like not trivial to review, please help to make it easier by the following steps:

- break it up into several commits, each of them testable on its own. If these would depend on each other, please start with a single commit of the root matter only
- please look at TravisCI result. Ignore the miss-configured run that only went about a minute or two. The serious runs (> 20 min runtime) show some broken tests. Ignore test_urllib2net, according to http://bugs.jython.org/issue2708. Can you reproduce the other broken tests locally with ``ant regrtests``? All broken tests need to be resolved before we can eventually merge this. I would prefer not to attempt a serious review before this is somehow fixed or at least understood. It might require some redesign that would then double the workload of reviewing.
I however believe that splitting the commit will also make fixing the broken tests much easier.
- can you add possibly minimal code examples that demonstrate the issues fixed? Even better, write them as tests.
- while there are bugfixes that do not strictly require tests it is always very preferable to have tests for newly fixed issues, last but not least to avoid later regressions. Do you think it would be feasible to write tests for this matter?
- If issues have bigger implications or involve larger design fixes, they should be tracked on bugs.jython.org. Do you think this might be the case for some of these issues?

In your description, I assume you mean [CPython](https://github.com/python/cpython) instead of [Cython](http://cython.org/)...? (It's better to avoid mixing these up)

jeff5:
Although the explanation is long, the patch isn't. Maybe I'm seeing already broken up? (E.g. the note addresses str/repr confusion which is the subject of #113.) Generally, though, I welcome this. Thanks for exploring it @devendor .

A number of hazards arise from the fact that ``PyUnicode`` extends ``PyString``. This is catching some more places where we have to have special logic, maybe? The way to make ``memoryview(u"hello")`` raise a type error, in my view, is to override ``getBuffer`` in ``PyUnicode`` so it raises one. I think that's the fundamental omission.

I'm not surprised there are a few rough edges in bytearray as it was my first big contribution.

It is clear looking at struct.java that the design is very old. ByteStream looks so like a NIO ByteBuffer (and PyByteArray) that I can't help thinking it would benefit from a re-write. But going for minimal change is probably right for now. I'll add further comments to the PR. Apologies in advance for any clumsiness with those tools (still practising).

src/org/python/core/PyByteArray.java
```diff
@@ -1325,8 +1325,10 @@ final synchronized PyObject bytearray___iadd__(PyObject o) {
         } else if (oType == PyString.TYPE) {
             // Will fail if somehow not 8-bit clean
             setslice(size, size, 1, (PyString)o);
+        } else if (o instanceof PyMemoryView){
```
jeff5: Couldn't we accept here any object implementing the buffer interface?

src/org/python/core/PyMemoryView.java
```diff
@@ -67,7 +67,8 @@ static PyObject memoryview_new(PyNewWrapper new_, boolean init, PyType subtype,
         ArgParser ap = new ArgParser("memoryview", args, keywords, "object");
         PyObject obj = ap.getPyObject(0);
 
-        if (obj instanceof BufferProtocol) {
+        // memoryview in 2.7 throws typeerror on unicode rather than assuming encoding like buffer().
```
jeff5: Not this way. It's important that we accept any (byte) buffer object. Try ``n = memoryview(buffer(b'abc'))``. Solve by defining ``PyUnicode.getBuffer()`` to throw?

src/org/python/modules/struct.java
```diff
@@ -1096,11 +1097,36 @@ public static PyTuple unpack(String format, PyArray buffer) {
             throw StructError("unpack str size does not match format");
          return unpack(f, size, format, new ByteStream(string));
     }
-    
+
```
jeff5: Again, I think we should accept any byte-like object (``BufferProtocol``) here but I haven't looked at the other ``unpack`` methods. It's a pain we have to make a string copy just so we can put the bottom 8 bits of each ``char`` in the ``ByteStream``. (Do we?) If it doesn't provoke a cascade of changes elsewhere, could we teach ``ByteStream`` the buffer protocol so it gets the bytes directly?

Stewori:
> Although the explanation is long, the patch isn't.

I honestly did not really look into the code because it breaks six tests, ignoring test_urllib2net.

```
     [exec] 7 tests failed:
     [exec]     test_array test_bytes test_bytes_jy test_gzip test_memoryview
     [exec]     test_types test_urllib2net
```

I think a divide and conquer approach would be best to isolate the breakages before further review. And the cumulative nature suggests natural splitting points, allowing to sooner accept those parts without breakage.

devendor:
@Stewori @jeff5 I'm working through quite a few of these and think I've found a test that isn't testing what it thinks it's testing.

From Lib/tests/test_str_jy.py line 33
```
code_resistance(self):
        # Issue 2037: prevent byte/str elements > 255
        self.assertRaises(UnicodeEncodeError, str, java.lang.String(u"caf\xe9 noir"))
        self.assertRaises(UnicodeEncodeError, str, java.lang.String(u"abc\u0111efgh"))
```
Now `\xe9` is decimal 239 and while `\u0111` is greater than 255, the code in PyString that checks for byte values throws IllegalArgumentException and not UnicodeDecodeError.

With my current tree, I can get some UnicodeDecodeError results in a console if I set PYTHONIOENCODING=ascii and allow the console to echo values rather than assigning them.  As near as I can tell, it seems like ArgParser might have been trying to decode arguments and grammers as ascii but seems to be taking bytes in my current tree.

I think maybe redirecting toString has fixed some issues in lex parsing when byte values are > 128, ie: not ascii.

jeff5:
@devendor The way the parser handles strings is certainly counter-intuitive but I don't think it's the problem here. Source code should be ASCII unless there is an encoding declaration ([PEP 263](https://www.python.org/dev/peps/pep-0263/)).

The comment is wrong (should be >127), and poor since what is really tested here is that implicit encoding to bytes during construction enforces ASCII range, and neither string is acceptable. The ``UnicodeDecodeError`` is raised during [new](https://github.com/jythontools/jython/blob/master/src/org/python/core/PyString.java#L175), not the Java constructor. However, we've had ample evidence since then that the test in the constructor raises ``IllegalArgumentException``. As predicted in the related issue [bjo2037](http://bugs.jython.org/issue2037) enforcing the byte nature of ``str``, when it was cleverly UTF-16 originally, exposed a good deal of usage that has given us pain ever since, ultimately for good I think.

Reference behaviour is:
```
>>> import java
>>> s1 = java.lang.String(u"caf\xe9 noir")
>>> type(s1)
<type 'java.lang.String'>
>>> s1.charAt(3)
'\xe9'
>>> str(s1)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 3: ordinal not in range(128)
```
The wrapping in java.lang.String may be pointless as it ends up a ``PyUnicode`` anyway.

``PYTHONIOENCODING`` should have nothing to do with this. You may just be seeing unpleasant side effects that look similar, as Jython tries to format an error message containing Unicode. Lexing the source is ok AFAIK, but I've doubts about the reverse process of giving error context when not ASCII. 

``sys.setdefaultencoding`` would affect what you're studying, but is discouraged:
```
>>> import sys
>>> reload(sys)
<module 'sys' (built-in)>
>>> sys.setdefaultencoding('utf-8')
>>> str(s1)
'caf\xc3\xa9 noir'
```

devendor:
@jeff5 @Stewori 
I haven't checked in the branch yet, but I have all tests succeeding with all of the changes applied and a very small change to allow unwrapped java.lang.String to pass through a lower bar of 8bit clean which reflects the strange way that py2 treats byte IO and strings as synonymous with a 7bit default ascii encoding.

The test is below demonstrates the new behavior.

```
    def test_unicode_resistance(self):
        # Issue 2037: prevent byte/str elements > 255
        # Enforce 7 bit clean int(char) < 128 on PyUnicode strings as py2 does.
        self.assertRaises(UnicodeEncodeError, str, u'caf\xe9 noir')
        # Allow 8 bit clean int(char) < 256 in java.lang.Strings when not specifically
        # subclassed as PyUnicode or otherwise packed into a PyString. Replicates
        # py2.x ambiuguity of using bytes and strings synonymously with default ascii
        # encoding and a disregard for non-printable bytes.
        self.assertEquals(str("caf\xe8 noir"), str(java.lang.String(u"caf\u00e8 noir")))
        # Treat as PyUnicode if java.lang.String codepoints > 255 even when not
        # subclassed as PyUnicode.
        self.assertRaises(UnicodeEncodeError, str, java.lang.String(u"abc\u0111efgh"))
        # Don't be more lazy than python with str/byte ambiguity. Only unwrapped
        # java.lang.String gets a bye on ascii 7bit clean since it is likely to contain
        # byte driven IO based on how python treates str type in py2. Otherwise we'll
        # enforce strong typing and explicit decode/encode if you want something else.
        class test7BitDirtyJavaObject(java2py(java.lang.Object)):
            def toString(self):
                return java.lang.String(u'caf\xe9 noir')
        class test7BitCleanJavaObject(java2py(java.lang.Object)):
            def toString(self):
                return java.lang.String(u'cafe noir')
        self.assertRaises(UnicodeEncodeError, str, test7BitDirtyJavaObject())
        self.assertEquals(str("cafe noir"), str(test7BitCleanJavaObject()))

```

devendor:
Closing this one as it is broken up in other pull requests now.

PR #113, opened 2018-10-13
Str  recursion

devendor:
This long story starts 11 years ago in the initial implimentation of @ExposedMethod.

git diff  b06512177^  b06512177 -- src/org/python/core/PyObject.java

When you @Expose __str__ and then override it in Py...Derived by the same name, the Py...Derived method ends up in the exposed __str__() method tree, so that a Type().lookup("__str__") within the down stream recursive method will find itself, but behind an ExposedMethod veneer that prevent it from realizing the recursion.  When this error is left in place, any derived class that doesn't state a __str__ method will fail with a max recursion stack error.

This way exposes object___str__ as __str__, so that derrived __str__ methods are not direct overrides and don't wind up in the lookup table for __str__.  The same is true for __repr__.

An alternate solution that makes the problem more obvious looks something like this in Py...Derived.
```

    public PyString __str__() {
        PyType self_type=getType();
        PyObject impl=self_type.lookup("__str__");
        PyObject my_impl=TYPE.lookup("__str__");
        if (impl != null && !(my_impl == impl)){
               ....
         }
         return super.__str__();
     }

```
Anyway... the initial mistake lead to a recursion issue that was quickly forgotten behind an ugly workaround that mapped __str__ to __repr__ and __repr__ to toString() and toString to (String)__repr__ in Py...Derived.

When reviewing this, PyObject.java and src/templates/gderived-defs are the two main changes.  The rest are autogenerated defs recreated via cpython src/templates/gderived.py, and just contain the change to map toString() back to toString() intelligently.  This gives us back the straight forward granularity of toString for java-ish (String)s, __str__ for PyString (str)s and __repr__ for PyString (repr)s which most of the time can be almost the same thing, but sometimes it matters like with bytearray.

This will need some testing and maybe some downstream cleanup since there is 11 years of workarounds down stream, but it seems to fix more than it breaks with the limited initial look at it I've given.


Stewori:
@devendor This looks like not trivial to review and I am not sure if I entirely understand the issue that is fixed. Is it tracked in some form at bugs.jython.org? If not it would be worth tracking there. Fixing an eleven year old workaround can have further implications that might require more extensive discussion. bugs.jython.org is the suitable place for this.

To help me understand the issue better, would it be possible to add possibly minimal code examples that demonstrate the issue fixed?

That said, I copy/pasted two points from my comment on #112 that apply here as well:

- please look at TravisCI result. Ignore the miss-configured run that only went about a minute or two. The serious runs (> 20 min runtime) show some broken tests. Ignore test_urllib2net, according to http://bugs.jython.org/issue2708. Can you reproduce the other broken tests locally with ``ant regrtests``? All broken tests need to be resolved before we can eventually merge this. I would prefer not to attempt a serious review before this is somehow fixed or at least understood. It might require some redesign that would then double the workload of reviewing.
- while there are bugfixes that do not strictly require tests it is always very preferable to have tests for newly fixed issues, last but not least to avoid later regressions. Do you think it would be feasible to write tests for this matter?

devendor:
I need to get the testing setup locally and get the various issues into a better state for merging.  I'll probably focus on the gzip, struct, bytearray, buffer, and memory view patches first as those are more straight forward.

In particular, this str and repr 11 year old work around thing is non-trivial.  I need to prioritize some work that pays for a bit, but will definitely circle back.  This feels like something that needs more experienced eyes and hopefully helps point someone with better context and stronger java skills at this corner of the code.

From a debugging and reproducing the initial issue, you can just take the master tree with some slight modifications to set exposedmethods for any simple version of __repr__ and __str__ directly as seemed to be the case prior to the work around.  From there you can do something like the following to reproduce an infinite recursion.

`
class myObj(object): pass

myObj().__str__()

`
Then add some break points in PyObjectDerived.__str__ to catch infinite recursion in action.  Repr is a little stranger when troubleshooting interactively as you can wind up calling that implicitly as the interactive shell tries to display what you are doing if you don't assign the return object, but it's the same sort of thing.  Derived objects that override __repr__ and __str__ work fine, and non-deirived objects work fine. but any derived object that doesn't override the exposed method falls into infinite recursion.

I'm not sure this is the right fix, but it seems like it might be a step in the right direction.  I think I bumped into it and dug in out of obsessive tenacity, but it's pretty deep in the code base and stretches my own sense of where my skills lie both in terms of java and a code base that I'm new to.  

If you have any time to dig into one of the three, it would be great to do some discovery here while I get the other bits broken up with unit testing.

The compromise in the workaround seems to be polluting what was originally a fairly discrete difference between use of toString() to produce java compatible String objects, __repr__ for repr PyStrings, and __str__ for __str__ Pystrings.  In most cases it might not have mattered too much since the __repr__ output isn't used as a dependency in most cases, and many toString() and __str__() methods are nearly the same with the exception of casting to PyString, but when you dig into code points, byte arrays, and decoding thing inside of struct, the murkiness and confusion of the initial work around gets really uncomfortable.  Also, when you back out the workaround, several other things like faithfully reproduced output for buffer().__repr__() come alive in the old code base to replace the generic output of the PyObject final exported method.

Thank you for the feedback and help. Sorry for pitching this before it was ready.

devendor:
This version takes a slightly different approach and protects against the problem in defs.  This one should be ready for more of a review.

For review, following this order will make things more clear.

1. pull in Lib/test and run them without any other changes.
2. pull in src/org/python/core/PyObject.java and Py2kBuffer.java
3. build and re-run. this solves one problem and causes others.
4. Pull in the src/org/python/core files: JavaProxyMap.java PyJavaType ThreadState.java PyInstance.java
3. Build / test.This should fix the nested recursion issue in the new test_dict.py tests and the py_string_jy issues.
4. At this point you can pull in BaseDictionary BaseSet, PyDictionary, PyList and PyStringMap.  Just take a look and note that this is an extremely minor change to wrap a try: finally: around repr thread state recursion object in ThreadState.
5. The rest are just autogenerated classes.  Pull in src/templates/gderived-defs.
6. cd to src/templates and run /path/to/cpython gderived.py
7. You should be able to git diff my-branch --stat and see that the auto generated files match what you just recreated.
8. rebuild and test again.

jeff5:
I'm happy to take a look at this as [this](https://github.com/jythontools/jython/blob/118aa8c612c021ca7bb523dfdedf21ae45706ed9/src/org/python/core/PyObject.java#L223-L231) has always made me feel what I was doing with `__str__` etc. was unsafe. I can see it will take the weekend, at least.

Lib/test/test_derived_stack_jy.py
```diff
@@ -0,0 +1,53 @@
...
+        # __repr__ is exposed
+        self.assertEqual(b.__repr__(),repr(b))
+
+    def test_derived_recursion(self):
+        """when PyObject exposes a method by it's actual name, the exposed facade delegates through
+        "  subclasses in PyObjectDerived and other ..Derived.  This fixes the above test, but opens
+        "  up a recursion issue.
+        """
+
+        class F(object):
```
jeff5: Nice test. Because `__str__` and `__repr__` cross-refer I wanted to check what happens if you only override one of them, so extending your test I wrote:
```
        class E(object):
            """This never passes through to super.__str__ or super.__repr__ in PyObjectDerived.java"""
            pass

        class F(object):
            def __repr__(self): return "myrepr"

        class G(object):
            def __str__(self): return "mystr"

        class H(object):
            def __repr__(self): return "myrepr"
            def __str__(self): return "mystr"

        for C in (E, F): # Both have original object.__str__
            C().__str__()
            str(C())

        for C in (E, G): # Both have original object.__repr__
            C().__repr__()
            repr(C())

        for C in (G, H): # Both override __str__
            self.assertEqual(C().__str__(), "mystr")
            self.assertEqual(str(C()), "mystr")

        for C in (F, H): # Both override __repr__
            self.assertEqual(C().__repr__(), "myrepr")
            self.assertEqual(repr(C()), "myrepr")
```

Lib/test/test_derived_stack_jy.py
```diff
@@ -0,0 +1,53 @@
...
+            """Doesn't trigger recursion when exposed method is overriden."""
+            def __repr__(self): return "myrepr"
+            def __str__(self): return "mystr"
+
+        class E(object):
+            """This never passes through to super.__str__ or super.__repr__ in PyObjectDerived.java"""
```
jeff5: Never or always?

src/org/python/core/PyObject.java
```diff
@@ -221,22 +221,44 @@ void proxyInit() {
...
+     * A good example of consequence is Py2kBuffer buffer() objects which should return the actual content
+     * byte string on __str__() not __repr__() and custom __repr__() on __repr__() rather than
+     * the generic object_toString.
      */
-    // XXX: more detail for this comment is needed.
-    @ExposedMethod(names = "__str__", doc = BuiltinDocs.object___str___doc)
-    public PyString __repr__() {
+    /**
+     * Equivalent to the standard Python __repr__ method. This method should not typically need to
+     * be overrriden. The easiest way to configure the string representation of a
```
jeff5: Is this still good advice?

src/org/python/core/PyObject.java
```diff
@@ -221,22 +221,44 @@ void proxyInit() {
      * <code>PyObject</code> is to override the standard Java <code>toString</code> method.
      **/
     /*
+     * This was as exposed __str__ method in the original @ExposedMethod annotation
```
jeff5: I would not record the history here. Comments should relate to the current state (which since it is "intutive" could be concise).

src/org/python/core/PyStringMap.java
```diff
@@ -240,25 +240,28 @@ final String stringmap_toString() {
...
-        for (Entry<Object, PyObject> entry : table.entrySet()) {
-            Object key = entry.getKey();
-            if (key instanceof String) {
-                /* This is a bit complicated, but prevents us to duplicate
-                   PyString#__repr__ logic here. */
-                buf.append(new PyString((String)key).__repr__().toString());
-            } else {
-                buf.append(((PyObject)key).__repr__().toString());
+        StringBuilder buf = new StringBuilder( "{" );
+        try {
```
jeff5: Am I right in thinking the `try...finally` here is to protect the repr stack against bad implementations of `repr`? A thing we maybe should have had all along, rather than newly-necessary as part of this fix?

I'm not suggesting a separate PR, just trying to understand how it fits in.

src/org/python/core/ThreadState.java
```diff
@@ -43,14 +43,17 @@ public PySystemState getSystemState() {
...
+
     public boolean enterRepr(PyObject obj) {
         if (reprStack == null) {
             reprStack = new PyList(new PyObject[] {obj});
             return true;
         }
+        Boolean objHasTp = (obj instanceof Traverseproc);
         for (int i = reprStack.size() - 1; i >= 0; i--) {
-            if (obj == reprStack.pyget(i)) {
+            if (objHasTp && obj == reprStack.pyget(i)){
```
jeff5: I don't follow this. Looks plausible but a comment from @Stewori would reassure me. Again, is this maybe a thing that was lacking all along, not a consequence of this change?

src/templates/gderived-defs
```diff
@@ -156,13 +156,19 @@ define: (ClassBodyDeclarations)ctr_userdict
 define: (ClassBodyDeclarations)toString
   public String toString() {
     PyType self_type = getType();
-    PyObject impl = self_type.lookup("__repr__");
-    if (impl != null) {
-      PyObject res = impl.__get__(this,self_type).__call__();
-      if (!(res instanceof PyString))
+    PyObject impl = self_type.lookup("toString");
+    if (impl != null && impl != TYPE.lookup("toString")) {
+      Object res = impl.__get__(this,self_type).__call__();
```
jeff5: Can `__call__` really produce something other than a `PyObject`? I don't see how. And so some of these added clauses don't apply.

src/org/python/core/ThreadState.java
Stewori: Sorry I don't really see the relation between ``Traverseproc`` and ``JavaProxy``.
@devendor could you please elaborate what you intend to do here?
Why is checking the JavaProxy relevant if no Traverseproc is implemented?

jeff5:
@devendor These latest commits appear to be on the wrong PR.

jeff5:
@devendor This PR starts in a good way with coherent changes related to the subject, and you seem to have got to the bottom of a long-standing infelicity. This was welcome. I reviewed that (and @Stewori) and we made what I hope were constructive comments.

Since I downloaded the relevant work, you've pushed maybe a hundred changes that seem to have leaked from #123. It would be good if you would:
- revert this branch back to just what's relevant to the str-recursion problem.
- respond to the review comments on str-recursion.

The str-recursion fix seems basically good to me, and important enough that we should ask you to fill the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) to cover this and any other work on Jython (or any PSF project). I intend we look at all your Jython PRs eventually, but they deserve time for thought and discussion.

devendor:
@jeff5 Sorry for the mess.  I was monkeying with issues with tests and forgot what branch I was in.

devendor:
@jeff5 I've managed to annoy myself with this mess, so I started a new branch starting with current/master merged and validated by hand to make sure I didn't leave more garbage around.  I'm waiting on some validations right now.

devendor:
@jeff5  Pull request #126 reproduces this rebased off of current upstream master with two commits.  It should be more manageable and manually validating it seemed like a safer route to go given the mess I made in this branch.  Sorry about that.

jeff5:
Closing for the reasons @devendor explains in favour of #126. See also http://bugs.jython.org/issue2719.

PR #115, opened 2018-10-15
Zlib gzip testcase

devendor:
Test cases for the gzip bug.


devendor:
This is duplicated in the zlib fix branch.

PR #116, opened 2018-10-18
test_chdir fails on Windows filesystems that no longer support DOS shortname #2709

adamburkegh:
http://bugs.jython.org/issue2709

Bug summary: os.chdir(dosname) fails in test_chdir() because there is no dos shortname on some modern Windows filesystems, including, as it happens, my machine. Some other links on this behaviour in bug report.

The solution offered here is to call out to the windows command line to see whether a shortname exists.

I also add coverage for navigating to c:\Program Files\ as this is a system folder that, through standard or convention, usually created with a C:\PROGRA~1\ short name.

Have adapted the solution here
https://stackoverflow.com/questions/18893284/how-to-get-short-filenames-in-windows-using-java

... however could not get the inline version %~s1 working with the way jython treats command line parameters through the subprocess module, hence the batch file.

I tested the C:\Program Files\ behaviour but do not have a Windows filesystem with the 8dot3name property set differently to test against, unfortunately. 

The patch itself seems a bit fussy and ugly, but I guess it's also an edge case. Alternative suggestions of course welcome.


adamburkegh:
Thanks for the helpful comments @jeff5 . Have updated the PR to hopefully resolve.

jeff5:
Thanks @adamburkegh . I'll pull a patch and commit for you via hg. (I _think_ pressing "Approve" doesn't commit in gh. Otherwise I'm about to make a mess.)

jeff5:
Now in via hg at 60e4d43f920b8053f34c4d832267fa4d5453fa25. Thanks again.

PR #117, opened 2018-10-19
Tmp merge

devendor:
This is  a mega patch that includes all of the pieces I've been working on.  I need to break it up, but it's passing all tests including the new ones in this tree for me on linux.


devendor:
Alright not passing in the various automated testing tools attached to GitHub.  I might need some guidance to get those in line.

Stewori:
Indeed the automated CI's here are a bit special. This needs some improvement, since they're currently close to useless because of too many "regular" failures.

> I might need some guidance to get those in line.

They usually rerun the failed tests in verbose mode. Sometimes that helps to get a clue what's going wrong. I usually look at the runs that don't fail right at the beginning due to config issues (@darjus  maybe we'd want to just remove these...?). Then compare to the CI logs of the last commit or some few last commits to get an idea of what new breakages might have been added.

On bigger changes -- and this PR qualifies as such I suppose -- regrtests should also be checked on Windows (not covered by CI) and OSX (Travis CI reports 10 failures but I did not investigate if these already failed before), using Java 7 and 8 if possible. There can sometimes be more or less subtle differences in behaviors between platforms. I can see if I find time to run regrtests on Windows but don't promise it. I don't have OSX available for testing however.

devendor:
This one can be closed.  The str recursion one should be ready for review and isolated to recursion and the 11 year old __repr__ workaround exclusively.

devendor:
Found the close button.  This one is being broken up in other requests.

PR #119, opened 2018-10-24
memoryview should reject unicode input with TypeError

devendor:
Memoryview in py2.7 raises TypeError on unicode input whereas in jython, it strips Unicode encoding entirely. The older buffer interface encodes unicode as utf-16be.

Downstream code in pip, msgpack, and likely other places, catch TypeError and adjust for the Unicode input.

I'm not sure what the used case is for other array types that py2.7/memoryview doesn't support being supported by jython, but it seems like something that may have been added because the design allowed it.  These are not likely to typecast into something usable in downstream code and may also be more of a liability than a feature.  

Since I did not find cases where these cause breakage, and don't know who might be using memoryview of the other array types, I added code that would allow targeted removal of specific array types by removing the typecode from `accept_array_types = "zcbBhHiIlLduf"`.  There is also a commented out try catch block in the test_array.py code that would fix tests attempting to create a memoryview of the other types. Currently it includes all types so it has no effect other than blocking automatic addition of any new array types to the permitted list, and pointing the way to a quick fix if we discover this causes downstream issues like the Unicode memoryview does.

### DIFF
--- a/src/org/python/core/PyMemoryView.java
+++ b/src/org/python/core/PyMemoryView.java
@@ -67,8 +67,30 @@ public class PyMemoryView extends PySequence implements BufferProtocol, Traverse
         ArgParser ap = new ArgParser("memoryview", args, keywords, "object");
         PyObject obj = ap.getPyObject(0);
 
-        if (obj instanceof BufferProtocol) {
-            return new PyMemoryView((BufferProtocol)obj);
+
+        /*
+         * Memoryview in 2.7 throws typeerror on unicode rather than assuming encoding like buffer().
+         * Various other array derivitives whos content encodeing can't be verified do the same. Reliable typing
+         * is one of the features of memoryview over the older buffer Py2kbuffer, and throwing the right exception
+         * is used in downstream modules like msgpack and pip even if technically there is a buffer interface behind
+         * the target object in our instance.
+         *
+         * This still accepts basically all of the PyArray typed variants, but probably shouldn't. Adjusting type
+         * string can granularly remove them but we should make sure there is not an active
+         * use case for this extra functionality before we get rid of it, and optionally allow a flag
+         * to ignore the type check if we really need memoryviews of array.arrays.
+        */
+        String accept_array_types = "zcbBhHiIlLduf";
+        if (obj instanceof Py2kBuffer || obj instanceof PyMemoryView || obj instanceof PyByteArray ||
+            (
+                obj instanceof PyString  && !(obj instanceof PyUnicode)
+            ) || (
+               obj instanceof PyArray
+               && obj.__getattr__("typecode") != null
+               && accept_array_types.contains( obj.__getattr__("typecode").toString() )
+            )
+           ) {
+            return new PyMemoryView( (BufferProtocol) obj );
         } else {
             throw Py.TypeError("cannot make memory view because object does not have "
                     + "the buffer interface");
--- a/Lib/test/test_array.py
+++ b/Lib/test/test_array.py
@@ -873,6 +873,15 @@ class BaseTest(unittest.TestCase):
             a[1:-1] = array.array(self.typecode, self.example[1-n:-1])
 
         v = memoryview(a)
+
+        # PyMemoryView still currently accepts all inbound array types but can be
+        # made more restrictive via accept_array_types in PyMemoryView Py2kBuffer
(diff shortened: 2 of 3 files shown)


jeff5:
In CPython, `buffer`  is implemented on top of an old API. Later, a more capable buffer protocol was invented and `memoryview` was implemented on top of it. As I understand it, motivation for the new API was to allow `numpy` to adopt it in place of their own C API, and so any object implementing the modern buffer API would be accessible this way to `numpy`. For this reason, the modern buffer API supports multi-dimensional arrays and elements that are not just a byte.

Jython, which had intended never to implement the Python `buffer` type until this point, implemented (the nearest Java equivalent to) the modern buffer API and `memoryview` on top. Then because it was easy and seemed to be expected downstream, we implemented `buffer` also on top of the modern API.

However, `memoryview` on [`array`](https://docs.python.org/2/c-api/buffer.html) and [`unicode`](https://docs.python.org/2/library/stdtypes.html#memoryview) are excluded in Python 2.7, so I'm in favour of tidying this up. They're untested and `unicode` is/has broken. It's better in Python 3, but doesn't look like CPython 2.7 will ever catch up with us on this point :) so it's probably unhelpful to exceed the CPython behaviour.

That's the background, and my view, but I'll need to think about whether this is the way to correct it.

jeff5:
This is really two problems:
1. `unicode` is accepted by `memoryview`, which we should fix, and
2. `array.array` is accepted by memoryview, which I think we should reduce to only the 'b' (byte) type.

I'm going to review the `unicode`part of this with the expectation of accepting it after changes, but the `array.array` part is complicated so I have raised [issue 2715](http://bugs.jython.org/issue2715) on the bug tracker.

src/org/python/core/PyMemoryView.java
```diff
@@ -67,8 +67,30 @@ static PyObject memoryview_new(PyNewWrapper new_, boolean init, PyType subtype,
...
+         * is used in downstream modules like msgpack and pip even if technically there is a buffer interface behind
+         * the target object in our instance.
+         *
+         * This still accepts basically all of the PyArray typed variants, but probably shouldn't. Adjusting type
+         * string can granularly remove them but we should make sure there is not an active
+         * use case for this extra functionality before we get rid of it, and optionally allow a flag
+         * to ignore the type check if we really need memoryviews of array.arrays.
+        */
+        String accept_array_types = "zcbBhHiIlLduf";
+        if (obj instanceof Py2kBuffer || obj instanceof PyMemoryView || obj instanceof PyByteArray ||
```
jeff5: The idea of the buffer protocol is that you already know how to talk to objects of types you haven't met. Therefore, an attempt to enumerate the acceptable types is misguided. The aim is to accept any type implementing `BufferProtocol`, but sometimes we have to exclude types whose claim is spurious. (One day, `PyUnicode` will no longer extend `PyString`.) As mentioned, `PyArray` is to be addressed separately

Without PyArray, all we need here is
```
if (obj instanceof BufferProtocol && !(obj instanceof PyUnicode)) { ...
```

src/org/python/core/PyMemoryView.java
```diff
@@ -67,8 +67,30 @@ static PyObject memoryview_new(PyNewWrapper new_, boolean init, PyType subtype,
         ArgParser ap = new ArgParser("memoryview", args, keywords, "object");
         PyObject obj = ap.getPyObject(0);
 
-        if (obj instanceof BufferProtocol) {
-            return new PyMemoryView((BufferProtocol)obj);
+
```
jeff5: I would say the comment is not needed, except to say that `PyUnicode` doesn't really support the buffer protocol. The PyArray bit will out of scope. We would only give background (e.g. about pip and msgpack) to explain a _non_-standard or unexpected choice. As you've pointed out, our motive here is conformance to expectations.

Also, it was too wide. You may find the [coding guidance](https://wiki.python.org/jython/JythonDeveloperGuide#Coding_guidance) useful.

jeff5:
As I looked into what to do about `array.array`, it seemed best to produce a unified solution for both that and `unicode`, now at: https://hg.python.org/jython/rev/616f1bbd4ec9 . It sprawls rather, but in many ways it is how things ought to have been in the first place.

Effectively, I've treated this PR a bug report. Thanks for pointing this out @devendor .

PR #121, opened 2018-10-24
Allow unpack and unpack_from bytearray and <type buffer>

devendor:
cstruct unpack in py2.7 allows unpack from <type buffer> and in 2.7.6 it adds support to unpack from bytearray. These features are used in msgpack/pip and likely other python modules.

This patch adds support for to unpack from the additional types and adds test cases based on tests/test_struct.py which is not run by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,14 @@ notifications:
 install:
   - ant
 
-language: java
-
 os:
   - linux
   - osx
+  - windows
+
+language: java
+ 
+dist: percise
 
 env:
   matrix:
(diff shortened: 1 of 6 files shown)


jeff5:
This seems basically a correct idea. It has acquired some noise form work on the Travis build, but I think I can filter that out and apply as a patch, if no-one objects to the slight forgery.

jeff5:
A trimmed version of this PR is now in at https://github.com/jythontools/jython/commit/bf70f7fba06cbbf23bf241927ccececd3a31f83b. Thanks Ray.

jeff5:
Raised as jythontools/jython issue 121

PR #122, opened 2018-10-24
This allows java String not already cast as a PyType to return an 8 bit

devendor:
clean PyString when possible instead of holding it to the higher bar of
7 bit clean by passing a PyUnicode to the upstream builtin.str.

The 8 bit permissive behavior is restricted to __str__ called on
java.lang.String object specifically, and reflects both the expected
return value from str and the ambiguity between byte values and string
in py2.7.

This seemed like a sensible change but was split out of the str / repr
recursion patch set as it is a behavior change not directly needed to
achieve the goals of the other patch set.

### DIFF
--- a/Lib/test/test_str_jy.py
+++ b/Lib/test/test_str_jy.py
@@ -2,6 +2,7 @@
 from test import test_support
 import java.lang
 import unittest
+from org.python.core.Py import java2py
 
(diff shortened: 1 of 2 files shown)


jeff5:
Under what circumstances is this useful?

Is the problem that sometimes what ought to be byte data emerges into Python as an abused `java.lang.String`?

PR #123, opened 2018-10-26
Fixes for travis tests.

devendor:
The various urllib2net ftp based tests fail due to nat in the travis environment.  Switched to a local twisted ftp server test on linux and skipped the ftp test in osx to fix it.

Newer xcode bundles java 10.  Pinned it to 9.3 for now until we get newer java working.

Travis trusty images (default) don't include oracle7jdk as it is no longer supported upstream by oracle.  Removed that test.  Openjdk7 still works.


devendor:
So I've been picking away to get a good set of tests.  Travis has some issues in that the newest linux platform is trusty which is 4 years old already, and oracle dropped version 7 so that's not even included in the platform for testing.  Openjdk7 is.  The osx tests were on the other side of the fence.  Xcode newer than 9.3 uses java10.  We can actually test windows 10 server on travis, but I had to speed up tests to complete in the 50 minute hard limit on jobs in the free tier.  This lead to some tweaking for parallel testing and general test performance improvements across the board.  The gc tests did show some sort of race when run in parallel.

Appveyor gives us 16.04 and 18.04 as well as windows 2012r2, and 2015 server with a ton of jdk and platform options.  Some tests are still failing in some combinations, but I think they're legitimate failed tests in some combinations of platform and jvm.  Openjdk7 on xenial fails two test_datetime_jy tests consistently for instance, and x86 oracle jdk8se on x64 platform fails in strange ways.  All of the windows builds fail two or three tests consistently as well.

This is looking like a good start for improving CI testing.  Note that in this version appveyor will upload test results so you can get detail from the test tab.  It also allows setting fast fail targets which I've done for xenial jdk8 and bionic jdk8 which pass all tests consistently.  In other words, if someone pushes something that fails the known good combinations at the top of the matrix, the rest of the matrix doesn't run.  This should prevent flooding the veyorbelt like I've probably done while finding my way.  As we get other combos working, toggling require_pass: true will allow them to become fast_fail targets as well.

jeff5:
@devendor This is producing good results I think. You have a clean run from Travis, which is welcome. I notice that although Appveyor shows green, it actually fails about half way through, so that would be misleading at a glance.

Is there any way you could test more carefully yourself before pushing?

I am concerned these scripts are getting more complex (paralell running, additional directories, fronting of high-risk tests or managing them according to length. I think this is trying too hard and favour something simpler to maintain for those that come after. However, these were set up oiriginally by @darjus and I'd respect his view above mine.

Also, I don't like the way CI-specifics intrude into test_urllib2net. It may be best simply not to enable the network resource (-u switch). Something that abstracts it more, at least.

jamesmudd:
I think this is an important issue that we do need to sort out the CI to be reliable. I'm not really happy with the current setup, and have been looking at Azure pipelines its free for open source and supports Linux/Mac/Windows builds. I have a working example at https://jamesmudd.visualstudio.com/jython/_build so you can see the interface, I think its better than anything we have at the moment. The config is in two files in here https://github.com/jamesmudd/jython/tree/azure-pipelines

A related point @jeff5 something weird is happening with the `ant regrtest` job some times the return code is `0` even thought there are failures causing CI to mark the build as good. Azure pipelines has a nice interface for looking at the test results.

devendor:
@jamesmudd @jeff5 I'm going to close this pull and keep this in my own repo until it's ready.

One thing of note is the strange exception that seems to happen in cleanup.  https://github.com/jythontools/jython/issues/124

Most of my assumptions in there are now in question and I've crawled completely back to being at a loss to explain it or even knowing if my suspicion that they're related is even correct. However...  I do know that parallel testing tends to bring out some specific issues more often and the windows exceptions occur outside of the test and are apparently not effecting test success or exit code of the process despite the output to stderr.

Anyway... the point is there are some cases, at least in windows, where things don't fail because they regrtest marked it passed and ant exits zero even though something is throwing to stderr.

jeff5:
Given latest comment, closing this.

PR #125, opened 2018-11-04
Locale, windows/cygwin and Java version failures in test_os_jy#2714

adamburkegh:
Mainly Turkish and Japanese test cases.


jeff5:
Committing this on James' say-so (via Mercurial as usual). Now at https://hg.python.org/jython/rev/359e573ade00 . It does in fact get us one step closer to a passing bot.

Thanks @adamburkegh !

PR #126, opened 2018-11-14
New str recursion pull

devendor:
Clean version of what was originally pull request #113 Str recursion.

This starts with an updated upstream master c54ee8424783a2ae14aea511ed002a2f9f897134.  Commits all manual changes in 572c5749ddaa2f8d72315f39425f43f6811dff06, then generates the 153 src/org/python/.../*Derived java files from the updated gderived-defs and src/template/gderviced.py tool used to regenerate them.

### DIFF
--- a/src/org/python/core/PyObject.java
+++ b/src/org/python/core/PyObject.java
@@ -221,22 +221,44 @@ public class PyObject implements Serializable {
      * <code>PyObject</code> is to override the standard Java <code>toString</code> method.
      **/
     /*
+     * This was as exposed __str__ method in the original @ExposedMethod annotation
+     * implementation way back in '07 and it orphoned a bunch of intentional overrides
+     * that faithfully reproduce proper __str__ and __repr__ behavior down stream objects.
+     * and making jython dirverge from behavior reproduced from the target cython.
+     *
+     * git diff  b06512177^  b06512177 -- src/org/python/core/PyObject.java
+     *
+     * Someone looked at it years later and added this.
+     *
      * counter-intuitively exposing this as __str__, otherwise stack overflow occurs during
      * regression testing.
+     *
+     * Then many years later, somone added this.
+     *
+     * XXX: more detail for this comment is needed...
+     *
+     * AFAIK, this was a small mistake in 07 and maybe the attempt to just pull it out without avoiding
+     * recursion caused some issue in some test a long time ago.  Still need some testing, but it's better
+     * to fix what caused the test issue if it still occurs.
+     *
+     * A good example of consequence is Py2kBuffer buffer() objects which should return the actual content
+     * byte string on __str__() not __repr__() and custom __repr__() on __repr__() rather than
+     * the generic object_toString.
      */
-    // XXX: more detail for this comment is needed.
-    @ExposedMethod(names = "__str__", doc = BuiltinDocs.object___str___doc)
-    public PyString __repr__() {
+    /**
+     * Equivalent to the standard Python __repr__ method. This method should not typically need to
+     * be overrriden. The easiest way to configure the string representation of a
+     * <code>PyObject</code> is to override the standard Java <code>toString</code> method.
+     **/
+
+
+    @ExposedMethod(doc = BuiltinDocs.object___repr___doc)
+    public PyString __repr__(){
         return new PyString(toString());
     }
 
     @Override
     public String toString() {
-        return object_toString();
-    }
-
-    @ExposedMethod(names = "__repr__", doc = BuiltinDocs.object___repr___doc)
-    final String object_toString() {
         if (getType() == null) {
             return "unknown object";
         }
@@ -253,15 +275,17 @@ public class PyObject implements Serializable {
         return String.format("<%s object at %s>", name, Py.idstr(this));
     }
 
+
+
     /**
      * Equivalent to the standard Python __str__ method. This method should not typically need to be
      * overridden. The easiest way to configure the string representation of a <code>PyObject</code>
      * is to override the standard Java <code>toString</code> method.
      **/
-    public PyString __str__() {
+    @ExposedMethod(doc=BuiltinDocs.object___str___doc)
+    public PyString __str__(){
         return __repr__();
     }
-
     /**
      * PyObjects that implement <code>org.python.core.finalization.HasFinalizeTrigger</code> shall
      * implement this method via:<br>
@@ -494,6 +518,7 @@ public class PyObject implements Serializable {
         return __call__(arg0, arg1, arg2);
     }
 
+
     /**
      * A variant of the __call__ method with four arguments. The default behavior is to invoke
      * <code>__call__(args, keywords)</code> with the appropriate arguments. The only reason to
@@ -716,6 +741,7 @@ public class PyObject implements Serializable {
         }
         return ret;
     }
+
 
     /**
      * Equivalent to the standard Python __setitem__ method.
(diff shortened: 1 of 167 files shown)


src/org/python/core/PyObject.java
```diff
@@ -221,22 +221,44 @@ void proxyInit() {
      * <code>PyObject</code> is to override the standard Java <code>toString</code> method.
      **/
     /*
+     * This was as exposed __str__ method in the original @ExposedMethod annotation
```
jeff5: I wouldn't spend any words here on how it used to be and why it was bad.

src/org/python/core/ThreadState.java
```diff
@@ -43,14 +43,17 @@ public PySystemState getSystemState() {
...
         return systemState == null ? Py.defaultSystemState : systemState; 
     }
-    
+
     public boolean enterRepr(PyObject obj) {
         if (reprStack == null) {
             reprStack = new PyList(new PyObject[] {obj});
             return true;
         }
+        Boolean objHasTp = (obj instanceof Traverseproc);
```
jeff5: I remember this attracted a comment from @Stewori in #113 that ran:

> Sorry I don't really see the relation between Traverseproc and JavaProxy.
> @devendor could you please elaborate what you intend to do here?
> Why is checking the JavaProxy relevant if no Traverseproc is implemented?

src/templates/gderived-defs
```diff
@@ -156,13 +156,19 @@ define: (ClassBodyDeclarations)ctr_userdict
 define: (ClassBodyDeclarations)toString
   public String toString() {
     PyType self_type = getType();
-    PyObject impl = self_type.lookup("__repr__");
-    if (impl != null) {
-      PyObject res = impl.__get__(this,self_type).__call__();
-      if (!(res instanceof PyString))
+    PyObject impl = self_type.lookup("toString");
+    if (impl != null && impl != TYPE.lookup("toString")) {
+      Object res = impl.__get__(this,self_type).__call__();
```
jeff5: `PyObject.__get__` and `__call__` can only return a `PyObject`.

jeff5 (changes requested):
This is all relatively trivial to spot and fix. When I've had time to experiment with the proposed change I'll be able to address the substance of the problem, wher ethe real value lies in your work.

jeff5:
Here is some thinking aloud. Just considering `__repr__` for now, if one were to follow the conventional pattern, it would look like: 
```
    public PyString __repr__() {
        return object___repr__();
    }

    @ExposedMethod(names = "__repr__", doc = BuiltinDocs.object___repr___doc)
    final PyString object___repr__() {
        return new PyString(toString());
    }

    public String toString() {
        ...
    }
```
I'm breaking out the behaviour into `toString` so that the advice we give to Java programmers still holds, that "The easiest way to configure the string representation of a `PyObject` is to override the standard Java `toString` method". (That should perhaps be a `PyUnicode` wrap in `object___repr__`, since the range of `toString` is any `String`. `PyObject.toString` is safe, but not all potential overrides.)

In `PyObjectDerived` we disregard the advice and Java-override `__repr__` so that, in any given instance of `PyObjectDerived`, we can search in the class dictionary of its Python-type for a definition of `__repr__`. We will normally find either the exposure of `object___repr__` or a `__repr__` made in the in Python class of this instance. (Third possibility: `__repr__` has been removed from the dictionary and we revert to Java `super`.) This is how it works for all "slot functions", and it seems ok to me.

The other feature of `PyObjectDerived` at present (before this PR) is that it overrides `toString` so it also looks for a Python `__repr__`:
```
    public String toString() {
        PyType self_type=getType();
        PyObject impl=self_type.lookup("__repr__");
        if (impl!=null) {
            PyObject res=impl.__get__(this,self_type).__call__();
            if (!(res instanceof PyString))
                throw Py.TypeError("__repr__ returned non-string (type "+res.getType().fastGetName()+")");
            return((PyString)res).toString();
        }
        return super.toString();
    }
```
This supports the facility we give Python programmers to override Java `toString` from Python. However, this now gives us a stack overflow from:
```
>>> class C(object): pass
>>> c = C()
>>> repr(c)
```
because `object___repr__` calls `toString`, overidden by `PyObjectDerived.toString`, which finds the descriptor for the exposed `object___repr__`, which calls `toString` again.

I infer that this situation is what the unexplained cross-wiring of `__str__` and `__repr__` was supposed to fix, but being so confusing, even if technically correct (hard to be sure), it has been difficult to build upon it correctly. If we accept the archaeological theory, this PR is about how to solve that original problem, and follow through the consequences for sub-classes of `PyObject`, in a less confusing way.

It may be incorrect for `toString` to look for `__repr__`, rather than `__str__`, once the cross-wiring is removed. In most writing about Python and Java the strongest comparison is between Java ` toString` and Python `__str__`, for example as the go-to in string interpolation for print. However, that change won't solve the recusion problem since `__str__` should defer to `__repr__` when not explicitly defined.

I consider the facility to redefine Java `toString` from Python (by defining `__str__`) to be important, and may affect a lot of users. I'm not so sure the advice to override `toString` instead of `__str__` is more than a consequence of the cross-wired solution, and it mostly affects us inside the project. Drop that and I think a different and less problematic organisation is possible.

jeff5 (changes requested):
I'm reviewing without the final change set, believing that to be all generated code. A correct understanding of the way slot functions work is essential. This is a difficult area, and I am sure I and others have got this wrong in subclasses of `PyObject`, which may be part of the cumulative problem.

I wrote a long-ish post above trying to identify the problem we (someone) tried to solve in 2007. My reversing (?) of the earlier cross-wiring fix reproduces the 2007 problem. but I think there may be simpler ways forward.

- There is a definite fault in exposing `__str__` directly that makes me doubt the analysis behind the PR.
- I question the second lookup added to every operation in every a `*Derived` class. Can it work without?
- Additional tweaks (`try-finally` and the involvement of garbage traversal) cloud the picture for me.

jeff5:
I opened http://bugs.jython.org/issue2719 to track this. The discussion here and in #113 is helpful. There are some legitimate bugs (explicit in Issue 2719), and the root cause may well be the confusing situation around `__repr__`, `__str__` and `toString`. It would be good to fix that, but the PR doesn't seem quite right in its approach to this latter historic confusion.

jeff5:
Is the real problem here just this divergence in Jython?
```
buffer
  repr(a):      <read-only buffer for 0x6, size -1, offset 0 at 0x7>
  a.__repr__(): <buffer object at 0x7>
  str(a):       hello
  a.__str__():  <read-only buffer for 0x6, size -1, offset 0 at 0x7>
```
Compared with CPython:
```
buffer
  repr(a):      <read-only buffer for 0x000000000309BCD8, size -1, offset 0 at 0x0000000003045F80>
  a.__repr__(): <read-only buffer for 0x000000000309BCD8, size -1, offset 0 at 0x0000000003045F80>
  str(a):       hello
  a.__str__():  hello
```
I agree that this stems from confusion caused by a bizarre and unexplained cross-wiring of `__str__`and `__repr__` in `PyObject`. There *is* a way to reverse this, but this PR hasn't quite found it. I have experimented myself, but it involves a lot of change that might be breaking for users.

I think we should fix the divergence superficially and improve the documentation, but leave the cross-wiring in place for now. This doesn't preclude a "proper" fix, but means we can take our time over it, and have somewhere to retreat to.

jeff5:
Useful discussion. Fixed in the change set at reference.

PR #131, opened 2019-03-09
support unicode exception

yishenggudou:
support unicode jdbc Error Exception

### DIFF
--- a/src/com/ziclix/python/sql/zxJDBC.java
+++ b/src/com/ziclix/python/sql/zxJDBC.java
@@ -327,7 +327,7 @@ public class zxJDBC extends PyObject implements ClassDictInit {
      * @return PyException
      */
     public static PyException makeException(PyObject type, String msg) {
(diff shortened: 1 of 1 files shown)


jeff5:
That looks good to me. Thanks for spotting it.

I think this is too small to bother you for a contributor form.

Our master repo is still Mercurial, so we don't commit via GitHub, but as a patch, which I'll take care of.

The tools we use don't handle Chinese very well, although we try to do better in Jython itself. Is "Kui Seng" (or "Kuì Sēng (愧僧)") an acceptable representation of your name in ACKNOWLEDGMENTS and the repo? Or is that something entirely different?

yishenggudou:
Thank you for reply
you can use my github name

jeff5:
Done the right way, Mercurial turns out to be ok with Unicode. Now at https://github.com/jythontools/jython/commit/379c6fa9334ede2627567e60b3b8bb60da8d80a1, thanks again.

jeff5:
Raised as jythontools/jython pull 131

PR #134, opened 2019-04-28
Upgrade bundled pip and setuptool

jamesmudd:
- pip 9.0.1 → 19.1
- setuptools 28.8.0 → 41.0.1


jeff5:
My recollection is that pip/setuptools after the versions we bundle is broken for Jython on Windows. It was the familiar sys.platform/os.name mess, but with the additional twist that it has been beaten into submission for posix. It would be great to be proved wrong.

Stefan (@Stewori) gave us a means possibly to address it via `PyShadowString` that I think we should have the courage to use more. I tried it experimentally once, but ISTR you have to fix pip too (in which case one could just fix pip).

jamesmudd:
Ok I didn't test on Windows. I will have a look at that. Should probably look at Mac too. Some more work for this. It's not critical just though the bundled version was getting quite old and it seemed to work on Linux.

jamesmudd:
I have tried this on windows 10 and it seems to work ok. I can install packages and it seems to work ok. It also seems to work ok on macOS. Would be happy to look at a ticket or a specific example if there are issues or maybe via a combination of other fixes this now works?

Stewori:
In bjo I spontaneously can only find these issues concerning newer pip, e.g.
https://bugs.jython.org/issue2736
https://bugs.jython.org/issue2660
https://bugs.jython.org/issue2345

In https://bugs.jython.org/issue2721 @jeff5 comments that upgrading to a new pip would break.

@devendor once opened a PR to fix something pip-related: https://github.com/jythontools/jython/pull/112
However, the PR was complex and contained various fixes and we requested him to break it up for sake of more organized reviewing. Various subsequent PRs were discussed and resolved. I don't remember how this finally turned out regarding pip.

jamesmudd:
After a little more testing I have managed to encounter the error:
```
    TypeError: can't concat memoryview to bytearray
```
However this issue is already under investigation/fixing see https://github.com/jythontools/jython/pull/119 and https://github.com/jythontools/jython/pull/120 I will have a look at those fixes as well, but with regard to the other issues I think this is close to working. I installed `pytest` ok and will have a better look at the other bugs.

I wonder what testing we really have around this? Should we be running some kind of pip tests? Try to install the top 50 packages on PyPI? Also we should maybe work more with pip, to get Jython included in pips CI (https://github.com/pypa/pip/issues/3044 and linked) ?

jeff5:
Thanks @Stewori, I had looked myself for that last issue but not found it. I revisited what I did at hat time and tried the upgrade (with some irrelevant lines and escape sequences suppressed):
```
PS issue2721> pip -V
pip 9.0.1 from C:\Jython\2.7.1\Lib\site-packages (python 2.7)
PS issue2721> pip list
pip (9.0.1)
setuptools (28.8.0)
virtualenv (15.2.0)
You are using pip version 9.0.1, however version 19.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
PS issue2721> pip install --upgrade pip
Collecting pip
  Downloading https://files.pythonhosted.org/packages/f9/fb/863012b13912709c13cf5cfdbfb304fa6c727659d6290438e1a88df9d848/pip-19.1-py2.py3-none-any.whl (1.4MB)
    100% |████████████████████████████████| 1.4MB 211kB/s ta 0:00:01
Installing collected packages: pip
  Found existing installation: pip 9.0.1
    Uninstalling pip-9.0.1:
      Successfully uninstalled pip-9.0.1
Exception:
  File "C:\Jython\2.7.1\Lib\site-packages\pip\basecommand.py", line 215, in main
  File "C:\Jython\2.7.1\Lib\site-packages\pip\commands\install.py", line 338, in run
  File "C:\Jython\2.7.1\Lib\site-packages\pip\commands\install.py", line 338, in run
  File "C:\Jython\2.7.1\Lib\site-packages\pip\req\req_set.py", line 795, in install
  File "C:\Jython\2.7.1\Lib\site-packages\pip\req\req_install.py", line 767, in commit_uninstall
  File "C:\Jython\2.7.1\Lib\site-packages\pip\req\req_uninstall.py", line 142, in commit
  File "C:\Jython\2.7.1\Lib\site-packages\pip\_vendor\retrying.py", line 49, in wrapped_f
    return Retrying(*dargs, **dkw).call(f, *args, **kw)
  File "C:\Jython\2.7.1\Lib\site-packages\pip\_vendor\retrying.py", line 212, in call
    raise attempt.get()
  File "C:\Jython\2.7.1\Lib\site-packages\pip\_vendor\retrying.py", line 247, in get
    six.reraise(self.value[0], self.value[1], self.value[2])
  File "C:\Jython\2.7.1\Lib\site-packages\pip\_vendor\retrying.py", line 200, in call
    attempt = Attempt(fn(*args, **kwargs), attempt_number, False)
  File "C:\Jython\2.7.1\Lib\site-packages\pip\utils\__init__.py", line 101, in rmtree
  File "C:\Jython\2.7.1\Lib\shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "C:\Jython\2.7.1\Lib\shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "C:\Jython\2.7.1\Lib\shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "C:\Jython\2.7.1\Lib\shutil.py", line 252, in rmtree
    onerror(os.remove, fullname, sys.exc_info())
  File "C:\Jython\2.7.1\Lib\site-packages\pip\utils\__init__.py", line 114, in rmtree_errorhandler
OSError: unlink(): an unknown error occurred: C:\Users\Jeff\AppData\Local\Temp\pip-hgrjvm-uninstall\jython\2.7.1\bin\pip.exe
PS issue2721> pip list
Package    Version
---------- -------
pip        19.1
setuptools 28.8.0
virtualenv 15.2.0
PS issue2721> pip install yolk
Collecting yolk
  Downloading https://files.pythonhosted.org/packages/2b/c0/73510e50668b70f3fad25328c58d7a29a300a2e452058dfdcb7c7538af7b/yolk-0.4.3.tar.gz (86kB)
     |████████████████████████████████| 92kB 590kB/s ta 0:00:01
Requirement already satisfied: setuptools in c:\jython\2.7.1\lib\site-packages (from yolk) (28.8.0)
Installing collected packages: yolk
  Running setup.py install for yolk ... done
Successfully installed yolk-0.4.3
PS issue2721>
```
Although not exactly smooth (that ol' unlink() problem again) it actually seems to have worked. This is better than I remember it. This is on 2.7.1, of course.

@jamesmudd : thanks for testing further. Maybe I'm wrong about later `pip` versions, or maybe we fixed something, and this may be down to #119 (fixed) or #120 which I was leaving the OP the satisfaction of fixing it. I'll poke, but if there's no interest in a few days, and my advice makes sense to you, maybe you could fix it?

jamesmudd:
Ok I have added the fix for `bytearray += memoryview` to this and it now seems to work in all the cases I have tested.
I have used the installer to install on Linux/Mac and Windows. and on all the platforms used `pip 19.2` to install:
- yolk
- pytest
- urllib3
- virtualenv

All seem to work. I have also tested uninstalling and reinstallign in various combination and it seems to work for me. If anyone else wants to try it as well that would be great.

When I quickly looked at the tests it seems we might have drifted away from CPython a little on the tests. 
the [Jython `test_bytes.py`](https://github.com/jythontools/jython/blob/master/Lib/test/test_bytes.py) is quite different from [CPython `test_bytes.py`](https://github.com/python/cpython/blob/master/Lib/test/test_bytes.py) in addition CPython have a [`test_buffer.py`](https://github.com/python/cpython/blob/master/Lib/test/test_buffer.py) maybe that could be ported to Jython?

jeff5:
Thanks @jamesmudd , that's the right way to get `iadd` I think. I'll commit this, but intend to do it in two steps (`iadd` and `pip`-things) in case the latter turns out to be a problem for beta users.

We should only compare with the 2.7 branch of CPython.  (Occasionally I have borrowed from a 3.x test, where we exceed 2.7.) `test_bytes.py` is pretty close to the one in our `lib-python/2.7` but that could be updated. The right psychological moment for upgrading `lib-python` is after 2.7.2.

We do run `test_buffer` ... . Ah, the one in 3.7 is quite different. This depends on a test object (type) defined in a C extension that presents a fully-featured buffer interface. I've seen that before, but ISTR in 2.7 this is a stand-alone test. There are Java tests with the same objective, within the limits of our implementation, and maybe their intent would be better met running a test object from Python code.

jeff5:
From the `iadd` patch I get
```
======================================================================
ERROR: test_iconcat (__main__.ByteArrayTest)
----------------------------------------------------------------------
  File "C:\Users\Jeff\Documents\Eclipse-O\jython-jvm9\dist\Lib\test\test_bytes.py", line 698, in test_iconcat
    b += u""
  File "C:\Users\Jeff\Documents\Eclipse-O\jython-jvm9\dist\Lib\test\test_bytes.py", line 698, in test_iconcat
    b += u""
ClassCastException: java.lang.ClassCastException: 'unicode' does not support the buffer protocol

----------------------------------------------------------------------
```
Looks like this should get converted to a `TypeError`.

Stuff that should work, does 👍:
```
>>> a = bytearray('abc')
>>> a += memoryview('def')
>>> a
bytearray(b'abcdef')
>>> a += buffer('ghi')
>>> a
bytearray(b'abcdefghi')
```
and even the added value:
```
>>> import array
>>> a += array.array('B', 'jkl')
>>> a
bytearray(b'abcdefghijkl')
```

I think this needs its own PR, treating #120 as the issue it fixes. I know #120 *is* a PR, but the submitter is unresponsive to review comments, while your approach is what I was advocating. It nicely illustrates why the recommended process for anything remotely complex is to open an issue, then a PR. Wwe can thank @devendor for identifying the issue.

`test_bytes_jy.BufferOperationsTest` was meant to be about the actual `buffer` type, so slightly the wrong place. The failing test `test_bytes.ByteArrayTest.test_iconcat` is where this belongs, but as we'd like to keep `test_bytes.py` like its CPython counterpart, I think we need a `test_bytes_jy.ByteArrayTest` (but not extending `test_bytes.BaseBytes`), that represents additional tests of `bytearray`.

I left some unfinished business in `add`, that it would be good to fix at the same time:
```
>>> a = bytearray('abc')
>>> a + 'def'
bytearray(b'abcdef')
>>> a + memoryview('def')
TypeError: can't concat bytearray to memoryview
>>> a + buffer('def')
```
Now why doesn't that fail in regrtest? Another use for `test_bytes_jy.ByteArrayTest`.

jamesmudd:
@jeff5 Thanks for taking another look at this. I did the patch a bit quickly and think I need to take another look at the tests, really only did enough to get pip to work. Your also right that we should split it into 2, it is 2 commits but it would be better. Of course you right I was looking at the CPython 3.7 tests so probably not much to take from there. Hopefully will get chance to have another look this week.

jamesmudd:
The iadd patch is now a separate PR #137 that should be completed before this is revisited.

jeff5:
I built myself an installer from the first change set (having already committed #137) and tested independently on Windows. Looks good!

Whatever experience instilled my former pessimism about later versions of pip seems now to be outdated.

jeff5:
Now fixed at https://hg.python.org/jython/rev/2d62972ea150, thanks @jamesmudd .

PR #136, opened 2019-05-08
Travis CI: Refactor to add Windows and drop Oracle JDK7

cclauss:
* https://chocolatey.org/search?q=ant
* https://chocolatey.org/search?q=java


jeff5:
@cclauss: Thanks for looking at this. Our CI falls short of its potential. Please post when you think this is ready to review. I note you have already agreed to a PSF CLA.

cclauss:
Hi Jeff,  The core problem is proudcanadianeh/ChocoPackages#28 and while I have Python skills, I do not have Windows skills nor do I have a Windows box to test on.  We will see progress can be made on that issue but until that happens, this PR will not progress.  An alternative would be to shift to https://www.appveyor.com or https://circleci.com for the Windows testing.

Would it be of interest to run tests on macOS?

Are there other things that you would want the tests to be able to do?

jeff5:
@cclauss: I was wondering why Windows on Travis. There is an AppVeyor integration already (I see you spotted), but I don't think it was ever configured to work. There was a working macOS integration too. (CircleCI?)

I develop on a Windows desktop and it is helpful to have something run the tests on other platforms. For others in the project, it is the other way around.

I have no experience with CI: I didn't set this up, but wish it was in better health. If the rot can be reversed, it will still need careful investigation and exclusion of failing tests. I say "careful investigation" because simply silencing an alarm is a poor policy. OTOH, failures may stem from limitations of the environment. What we should do is that and raise an issue and maybe tailor the test with skips sensitive to the CI.

That way, green becomes the norm and red actually means something. That would be useful!

PR #137, opened 2019-05-08
#2768 Allow bytearray + and += to accept buffer protocol objects

jamesmudd:


jamesmudd:
Think I have reworked this into a nicer solution. The `__add__` method duplicates and then calls through to the `__iadd__` method. This is a nice simplification of the code and fixes issues with the buffer protocol, however it **might** be coming at a slight performance degradation. The previous implementation performed two system array copies to combine the buffers. But without benchmarks we should be careful of assuming too much... and its more important to be correct than fast. If others feel differently I can have another look.

src/org/python/core/PyByteArray.java
```diff
@@ -926,32 +926,13 @@ public PyObject __add__(PyObject o) {
...
-
-        } else {
-            // Unsuitable type
-            // XXX note reversed order relative to __iadd__ may be wrong, matches Python 2.7
-            throw ConcatenationTypeError(TYPE, o.getType());
-        }
-
-        return sum;
+        // Concatenate the other buffer
+        return sum.__iadd__(o);
```
jeff5: I like that this is shorter and puts the logic in one place, but as you, not the avoidable copy that results when += inevitably finds there is no room for `o`. Everywhere else, this class does its utmost to avoid needless copying in the belief that sometimes byte arrays are big.

For this, it is only necessary to get the allocated size of the sum `bytearray` right, then it can look like the current handling of `PyString`.
```
            // Allocate the right size bytearray and copy the first part into it.
            sum = new PyByteArray(size + o.__len__());
            System.arraycopy(storage, offset, sum.storage, sum.offset, size);
            // Append o as bytes (if we can).
            return sum.bytearray___iadd__(o);
```
I'm seldom sure whether it is more appropriate to call e.g. `__iadd__` or `bytearray___iadd__`. Do we want a subclass to be able to intercept the action? I tend to think not here, as it is an implementation detail.

All types of `o` that may be added (and many others) have a `__len__` that won't throw, but I don't think we can check, other than by calling it. So we need a `try`. If you `catch` a `PyException`, you call `throw ConcatenationTypeError(TYPE, o.getType())`, and if it doesn't you now know how big the sum will be, unless the type is one `bytearray___iadd__` will root out.

Or maybe a quick check on the class of `o` is enough? I *think* all types of `o` that may be added (and a few others) bear the BufferProtocol, and have an `o.__len__()` it would be safe to call.

src/org/python/core/PyByteArray.java
```diff
@@ -1343,6 +1324,8 @@ final synchronized PyObject bytearray___iadd__(PyObject o) {
         } else if (oType == PyString.TYPE) {
             // Will fail if somehow not 8-bit clean
             setslice(size, size, 1, (PyString)o);
+        } else if (o instanceof BufferProtocol && !(o instanceof PyUnicode)) {
```
jeff5: Suggest `if (setsliceFromBuffer(...)) { /* we're done "/ }. I notice I haven't documented the return on that properly.

jamesmudd: Ah yes I didn't spot that. Will add the doc too.

src/org/python/core/PyByteArray.java
jamesmudd: Yes your right, I actually had the summed size at one point and changed it. I also agree its probably better to call `__bytearray___iadd__` directly, I will change that too.

jamesmudd:
I have reworked this using `setsliceFromBuffer` and think it now avoids the performance regression, if anything this is now an improvement. The `__iadd__` case is simplified too.

src/org/python/core/PyByteArray.java
```diff
@@ -1335,17 +1326,14 @@ public PyObject __iadd__(PyObject o) {
...
     @ExposedMethod(type = MethodType.BINARY, doc = BuiltinDocs.bytearray___iadd___doc)
     final synchronized PyObject bytearray___iadd__(PyObject o) {
-        PyType oType = o.getType();
-        if (oType == TYPE) {
-            // Use the general method, specifying the crack at the end of the array.
-            // Note this deals with the case o==this.
-            setslice(size, size, 1, (BaseBytes)o);
-        } else if (oType == PyString.TYPE) {
-            // Will fail if somehow not 8-bit clean
-            setslice(size, size, 1, (PyString)o);
```
jeff5: Because `PyString` is implemented using Java `char`, taking a buffer on it involves copying the whole thing to a `byte[]`, so the special `setslice` for `PyString` is worth it.

src/org/python/core/PyByteArray.java
```diff
@@ -1335,17 +1326,14 @@ public PyObject __iadd__(PyObject o) {
 
     @ExposedMethod(type = MethodType.BINARY, doc = BuiltinDocs.bytearray___iadd___doc)
     final synchronized PyObject bytearray___iadd__(PyObject o) {
-        PyType oType = o.getType();
-        if (oType == TYPE) {
-            // Use the general method, specifying the crack at the end of the array.
-            // Note this deals with the case o==this.
-            setslice(size, size, 1, (BaseBytes)o);
```
jeff5: As the comment points out, as well as being an easy optimisation, this deals with the case `a += a`, which I think this can't. Although ... CPython makes the same error. (Well, **I** think it's an error.)

src/org/python/core/PyByteArray.java
```diff
@@ -926,32 +927,22 @@ public PyObject __add__(PyObject o) {
...
-            sum = new PyByteArray(size + os.__len__());
+        if (o instanceof BufferProtocol && !(o instanceof PyUnicode)) {
+            // Duplicate this buffer into a new PyByteArray large enough for both
+            int totalSize = size + o.__len__();
+            PyByteArray sum = new PyByteArray(totalSize);
             System.arraycopy(storage, offset, sum.storage, sum.offset, size);
-            sum.setslice(size, sum.size, 1, os);
 
+            // Concatenate the other buffer
+            if (!sum.setsliceFromBuffer(size, totalSize, 1, o)) {
```
jeff5: Your `__iadd__` idea was simpler and deals better with `PyString`.

src/org/python/core/PyByteArray.java
```diff
@@ -1335,17 +1326,14 @@ public PyObject __iadd__(PyObject o) {
...
     final synchronized PyObject bytearray___iadd__(PyObject o) {
-        PyType oType = o.getType();
-        if (oType == TYPE) {
-            // Use the general method, specifying the crack at the end of the array.
-            // Note this deals with the case o==this.
-            setslice(size, size, 1, (BaseBytes)o);
-        } else if (oType == PyString.TYPE) {
-            // Will fail if somehow not 8-bit clean
-            setslice(size, size, 1, (PyString)o);
+        if (o instanceof BufferProtocol && !(o instanceof PyUnicode)) {
```
jeff5: You just need `if setsliceFromBuffer(size, size, 1, o) { }` as the third possibility, with a comment so it doesn't l;ook like a typo. `setsliceFromBuffer` itself checks for `BufferProtocol` and `PyUnicode` will reject you taking the buffer. Compare existing use in `setslice(int,int,int,PyObject)`.

jamesmudd:
Had another go at this, think it looks better this time, let me know if not.

jeff5:
This looks good to me and I've committed it locally (pending regrtest & push).

jamesmudd:
@jeff5 Great. Thanks for looking at this. Once this is merged I will go back and look at pip updates, think that should be fairly straight forward now but just want to test it as well as I can.

jeff5:
Now fixed at https://hg.python.org/jython/rev/4c627b7d4612, thanks @jamesmudd .

PR #138, opened 2019-05-08
#2762 Upgrade Apache commons-compress to 1.18

jamesmudd:


jeff5:
Now in via Mercurial at: https://hg.python.org/jython/rev/37616d952bde

Thanks James

PR #141, opened 2019-05-16
Centralize references and doco for registry keys supported by jython

adamburkegh:
Centralize references and doco for registry keys supported by jython itself into a single constants class.

`python.division.warning`  The deprecation notice was added by this change, but follows the CPython approach.

`python.options.caseok`	The code check is still in place and still seems to match the description, but we are well past Jython 2.0. I haven't been able to dig up what Jython 2.0 behaviour was here and whether this property is therefore dead code.

Changes to the supplied registry file are minor in this change. The documentation might be generated or cut down in the future.

Discussion prompting this at https://github.com/jython/jython.github.io/pull/12#discussion_r279166543


src/org/python/core/RegistryKey.java
```diff
@@ -0,0 +1,297 @@
...
+ * String values {@code true} or {@code yes} for true, or {@code false} or 
+ * {@code no} for false.
+ *
+ */
+public class RegistryKey {
+
+    private RegistryKey(){}
+
+    /**
+     *
```
Stewori: Is there a reason why some (i.e. most) comments have a leading blank line and some don't?

src/org/python/core/RegistryKey.java
```diff
@@ -0,0 +1,297 @@
...
+     * Setting {@code python.cachedir.skip} to true disables the package scan 
+     * for the cachedir (as defined by {@code python.cachedir} or a default). 
+     * Please be aware that disabling this will break importing from 
+     * java packages.
+     */
+    public static final String PYTHON_CACHEDIR_SKIP = "python.cachedir.skip";
+
+    /**
+     *
+     * {@code python.cpython2} is the name of the CPython executable.
```
Stewori: Maybe this should specify that it must be 2.7, i.e. not s.5.x or earlier. This is important because of bytecode format.

src/org/python/core/RegistryKey.java
```diff
@@ -0,0 +1,297 @@
...
+
+    /**
+     * 
+     * If this {@code python.security.respectJavaAccessibility} is false, 
+     * and you are using a Java version before Java 9, then Jython can 
+     * access non-public fields, methods, and constructors. 
+     * Normally, Jython can only provide access to public members of classes. 
+     * This may be deprecated in the future due to Java changes 
+     * to accessibility from version 9. Boolean.
+     */
```
Stewori: The JVM options ``--permit-illegal-access`` and ``--illegal-access=warn`` could be hinted in this context.

Stewori:
This looks very good to me. I appreciate in what exhaustive manner you identified the occurrences of these constants in Jython. Only had minor remarks for the description texts.
I assume this passes tests and everything. Then it should be fine AFAICT.

src/org/python/core/RegistryKey.java
adamburkegh: I'm used to a style where there is a leading blank line. I agree it's better to be consistent. Looking it up now, it seems Oracle doesn't have it.

https://www.oracle.com/technetwork/articles/javase/index-137868.html

Perhaps it used to be a default in Elipse or something? Anyway I will make them consistent and remove the leading blank line, unless there is a strong preference otherwise or some other existing style guide.

adamburkegh:
I should probably flag that this is `org.python.core.RegistryKey` rather than `org.python.RegistryKey` suggested by @jeff5. It just seemed lonely up at that level with only `Version.java` for company.

src/org/python/core/RegistryKey.java
adamburkegh: Added

src/org/python/core/RegistryKey.java
adamburkegh: Added.

adamburkegh:
@stewori Confirm tests pass. Technically they pass only as well as before - 5 windows fails for regrtest and 2 travis fails on ssh and ftp, Junit `org.python.indexer.IndexerTest` fails.

src/org/python/core/RegistryKey.java
jeff5: I find a space above and the line /** is all the blank I want. Thanks for the change.

jeff5:
I've committed this on my local repo via exported patches (squashed into one). I made some routine format adjustments. I will push when regrtest is done if I'm still at this desk.

Regarding the commit message, I try to stick to a guideline @fwierzbicki gave us a long time back: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
As a result I have the margin set in HgWorkbench at 72, but I keep forgetting the first (summary) paragraph should be less, which is what makes lists of commits readable in various lists. Here I am getting it wrong: https://github.com/jythontools/jython/commit/ef0ebf06f580a4d66330c81e5bc10d8708b7c5c7 . Looks like 69 is the limit here :( The recommended 50 seems needlessly short.

adamburkegh:
@jeff5 Cool. The commit message guidance makes sense, I'll try to stick with that.

jeff5:
Now in at https://hg.python.org/jython/rev/6e30f4eb4c99 , thanks Adam.

PR #142, opened 2019-05-20
FTP failures on Travis. Refresh latest test_urllib2net from CPython, with minor tweaks

adamburkegh:
This fixes the travis failure, Jython bug 2708, using the fix from CPython 17160. I'm not really a big fan of introducing direct references to travis in the test code, but I thought staying in synch with CPython was a greater good.

Did not attempt to synch the significant changes to test_support in this patch - it has moved to `test/support/__init__.py` in CPython 2.7 and 3.x latest. However it does move the two codebases closer together, incrementally.

Current CI state:
openjdk7, buffer overflow bug, unrelated
oracle jdk8, fixes test_urllib2net failures, test_ssl failure remains
OSX jdk8, fixes test_urllib2net failures, test_ssl failure remains

test_ssl failure looks like an expired certificate, but the code is more out of synch with CPython, and I wanted to review that diff as part finding it. This is submitted as a standalone patch which moves us one step closer to green, but doesn't quite get us over the line.


adamburkegh:
Oh one more question  - there is the copy of the python 2.7 libraries in `lib-python/2.7/test/test_urllib2net.py`. That copy is behind the CPython 2.7 branch. Does it get refreshed piecemeal, or does the whole folder get refreshed wholesale on some cycle?

Stewori:
Regarding ``lib-python`` see https://bugs.jython.org/issue2250. I once tried the update and found that it is significant work to get this right. So significant that I recommended it a won't fix for 2.7 at all.

adamburkegh:
@Stewori thanks for the background. I guess opportunistic merges like this one are the limit for now, then.

jeff5:
The answer to "does ` lib-python` get refreshed piecemeal or _en_ _masse_" is "usually _en_ _masse_". That of course has the effect of renewing a lot of our tests and is best done when we're willing to invite trouble.

Some of our tests, shadowing those in lib-python, are frozen and modified versions of their counterparts. Updating `lib-python` leaves them behind. Before writing a new test, I find it worth doing a diff with lib-python, or sometimes a more recent version, to see if I can borrow. If I can get rid of the shadowing file entirely, that's nice. We should do this too after an update of `lib-python`.

I may not have followed this discussion properly, but I think @Stewori is pointing us to a discussion about updating _en_ _masse_ (bad idea right now), while this patch is the sort of focussed update I advocate.

jeff5:
Now in at: https://hg.python.org/jython/rev/f3baccd1fc34 . [Two out of three](https://travis-ci.org/jythontools/jython/builds/551863760) is better than we've seen for a long time.

Thanks Adam. Et merci à @vstinner .

PR #143, opened 2019-05-23
test_ssl failures from weak certificate. Bug 2774

adamburkegh:
The minimal fix is simply the update to selfsigned_pythontestdotnet.pem
from Python core. See Python bug 36816.
https://bugs.python.org/issue36816

Also includes merges from latest 2.7 version of test_ssl, and separation
of most jython variation into test_ssl_jy.py. This is to ease future
maintenance and/or patches back into CPython. No new tests are skipped.
Method test_algorithms() is deleted as per 2.7 and Python bug 25764,
though it actually ran clean for me locally. https://bugs.python.org/issue25764

This PR is cousin to 
https://github.com/jythontools/jython/pull/136
https://github.com/jythontools/jython/pull/123
https://github.com/jythontools/jython/pull/142

Update: Travis CI has now run without any test_ssl failures.


jeff5:
Thanks for looking at this @adamburkegh. I poked around with the same certs last weekend when I finally understood why `ant regrtest` always failed at `test_ssl` (which passed when re-run at the prompt). This is a much more complete job.

jeff5:
Looks good to me. I moved a white space and added a news entry in your name. Apart from that, I'll push it via hg exactly as you submitted it. Thanks for this!

PR #140, opened 2019-05-14
Travis CI: Use declarative style

cclauss:
Exact same functionality but done in declarative style which is more compact and easier to understand.

PR #145, opened 2019-05-24
Hostname addon to fix openjdk7 buffer overflow. Issue 2651

adamburkegh:
In both the recent
https://travis-ci.org/jythontools/jython/jobs/535576706

and original 

https://travis-ci.org/jythontools/jython/jobs/313613197#L728

we see: 

/usr/lib/jvm/java-7-openjdk-amd64/jre/lib/amd64/libnet.so(Java_java_net_Inet4AddressImpl_getLocalHostName+0x1a5)

which suggests 
https://github.com/travis-ci/travis-ci/issues/5227

... is the culprit. Also, the FTP job failure shows the following:
0.00s$ hostname
travis-job-f506d6d7-075f-46e9-a40c-7245b966dcae

... which is pretty long.

The recommended fix is to use the hosts addon 
https://docs.travis-ci.com/user/hosts/

Darjus applied an early workaround to the travis config in c547fb5b96eaab78f0a8030a50f33ae68da85f04, but there is now a more standard one available, and maybe the old one doesn't work any more. 

CI has run, with expected number of failures and no buffer overflow crash in `test_builtin`. Since the issue is intermittent and depends on the hostname assigned, this is partial evidence. However the job log does show the hostname change being applied.

```
Set hostname to jyshort
$ sudo hostname jyshort
```


jeff5:
Thanks for tracking this down.

This was all greek to me, but now you identify it you can see how the implementation moves on from Java 7: http://hg.openjdk.java.net/jdk7/jdk7/jdk/file/9b8c96f96a0f/src/solaris/native/java/net/Inet4AddressImpl.c#l53 to something that looks a lot cleaner in Java 8: http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/solaris/native/java/net/Inet4AddressImpl.c#l86 .

jeff5:
I was perfectly happy to try this straight away, but now the build fails because it can't find `jdk_switcher`.
https://travis-ci.org/jythontools/jython/jobs/563244502

It didn't start with this push, but the one before:
https://travis-ci.org/jythontools/jython/jobs/563209523

while previously it worked fine:
https://travis-ci.org/jythontools/jython/jobs/561786472

Tempted to ignore it and see what future pushes produce..

jeff5:
A difference between failing and successful use of `jdk_switcher` is:

**Failing**:
```
Build dist: xenial
Build id: 563209522
Job id: 563209523
Runtime kernel version: 4.15.0-1028-gcp
```

**Successful**:
```
Build dist: trusty
Build id: 561786471
Job id: 561786472
Runtime kernel version: 4.4.0-101-generic
```

What's deciding that? What we are doing [seems to be recommended](https://docs.travis-ci.com/user/languages/java/), and not just for particular platforms.

adamburkegh:
This [thread](https://travis-ci.community/t/the-travis-ci-build-could-not-be-completed-due-to-an-error/1345/13) suggests jdks, especially older jdks, might need to be installed on xenial. We don't explicitly install the jdk, just ant, in our install step.

(I agree with the commenters that say this should be an implicit effect when using jdk_switcher, but maybe that's just how it works)

jeff5:
I've pushed this as a patch (clearly) but we don't seem to be quite there yet. It looks like `jdk_switcher` is only available on trusty. But we also have #140 that shows us how to do without it by declaring the matrix properly.

I feel the need of a bit of trial-and-error testing to merge #140 and what we have, but I'll do that on my own fork. It's time learned about this stuff.

PR #150, opened 2019-08-26
Remove accidental hg dependency

adamburkegh:
ant developer-build (the default) picked up an accidental hg dependency during rework to actually reduce hg/mercurial dependencies.

Without this change, was getting file not found in hg-metadata

```
hg-metadata:

BUILD FAILED
C:\Users\Adam\jython\jython7\jython\build.xml:472: Execute failed: java.io.IOException: Cannot run program "hg" (in directory "C:\Users\Adam\jython\jython7\jython"): CreateProcess error=2, The system cannot find the file specified
        at java.lang.ProcessBuilder.start(ProcessBuilder.java:1047)
        at java.lang.Runtime.exec(Runtime.java:617)
        at org.apache.tools.ant.taskdefs.launcher.Java13CommandLauncher.exec(Java13CommandLauncher.java:58)
```

Seems to be back to normal with this tweak.


jeff5:
Yes, sorry, you're right. I discovered it myself this afternoon and fixed it before I saw this. However, my solution is way more complicated than yours, so it must be preferable :)

I was actually trying to cover the case where there *is* an .hg directory, but no hg command. (I thought maybe the build servers would be like that.) But I've done both cases. Here you rely on the absence of .hg to tell you not to invoke hg, so I can still upset it by taking hg off my path.

adamburkegh:
Fair enough, glad it's sorted.

PR #151, opened 2019-08-27
Delete circle.yml

cclauss:
This is CircleCI v1.0 and I doubt it is worthwhile rewrite it for v2.


jeff5:
It's long gone, but thanks.

PR #154, opened 2019-09-30
CRLF win Java 11 lib2to3 test failure, bug 2808

adamburkegh:
This is a basic fix, reading the file using 'r' instead of 'rb'.
Checking in to separate fix from possible cleanup, and to test Unix on Travis
(local dev env is Windows).


adamburkegh:
Latest commit includes a merge with the latest CPython 2.x trunk for this file, plus overrides for features not available in Jython (https://github.com/python/cpython/commit/770a802df7057bcf0d72bbc01026fac8d396e1b3).

This fix passed locally on Windows 10, Java 8, 10 and 11. It makes Java 8 run clean and leaves Java 10 and 11 with one failing test, test_jython_initializer, that existed previously.

CPython 3 trunk seems to rely on new features more, particularly str() equality with no requirement for unicode(). CPython 2.x uses a callout to shell for diff and skips windows entirely, so the diff() code was already Jython specific and covers more than CPython for windows.

Recently we have followed the practice of introducing a _jy.py overriding test, and keeping the original as close as possible to CPython. This is a good technique. I have held off because lib2to3 has a dedicated directory of test cases populated by a custom __init__.py. So to introduce a new test involves also introducing an overriding __init__ that lists all the test cases by name. This seems heavyweight and risks missing future tests.

Happy to hear other opinions, as perhaps I am being over-cautious.

Lib/lib2to3/tests/test_parser.py
```diff
@@ -212,12 +215,33 @@ def test_multiline_str_literals(self):
...
 
+
 def diff(fn, result, encoding):
-    "A diff the result and original file content independent of OS."
+    """Diff the result and original file content independent of OS. This 
+    implementation is in Jython only. CPython 2.x calls out to the shell diff
+    command. CPython 3.x str equality behaviour is simpler and should allow
+    this test file to be replaced wholesale in Jython 3.x."""
     r = iter(result.encode(encoding).splitlines(True))
-    with open(fn, "rb") as f:
```
jeff5: Ah, ok. Here's the nub of it. r is a list of unicode but the file is being read as binary?

Lib/lib2to3/tests/test_parser.py
```diff
@@ -212,12 +215,33 @@ def test_multiline_str_literals(self):
...
+    lineNumber = 0
+    with open(fn, "r") as f:
         for line in f:
+            lineNumber += 1
             rline = next(r)
             if rline != line:
-                return True
+                return "original line {}:\n{}\n differs from:\n{}\n{}". \
+                       format(lineNumber,line,rline,diffLine(line,rline) )
```
jeff5: Comma-space, here and elsewhere. The CPython test suite is supposed to be PEP-8. We've been a bit lax, but some things are easy. (Perhaps I should use a linter.)

Lib/lib2to3/tests/test_parser.py
adamburkegh: In general it can be unicode, but actually the file it fails on just has ascii and CRLF.

Lib/lib2to3/tests/test_parser.py
adamburkegh: Will fix

adamburkegh:
@jeff5 Wish I knew why my machine is special : ) ... but I've been seeing the failure for quite a while, just put the locale stuff first. It's genuinely a test-only bug so I agree there's no reason to reopen the beta

Lib/lib2to3/tests/test_parser.py
jeff5: Oh hang on, looking properly and after coffee ... . This is an **encode**. So `result` should be `unicode` (or `bytes` to be interpreted as ASCII characters in Python 2) and `r` is definitely a sequence of `bytes` objects. So wasn't "rb" the right thing to do? Unless of course the encoding of what's in the file differs from `encoding`.

I notice in the CPython original that this test doesn't try to work on Windows, so it may not have dealt correctly with CRLF. If CRLF in the file is the problem, isn't the answer "rbU"?

Lib/lib2to3/tests/test_parser.py
adamburkegh: Fixed in latest for the few jython-specific lines. Rest I kept as character identical to merge from CPython

Lib/lib2to3/tests/test_parser.py
adamburkegh: That makes sense. I actually didn't know about file mode 'U'. In latest patch. Understood this will probably sit for a bit while the beta progresses.

jeff5:
Now in at https://hg.python.org/jython/rev/e81878891a4b. Thank you.

PR #155, opened 2019-10-07
NoSuchMethodError test_jython_initializer bug 2810

adamburkegh:
Fixes runtime errors where different return type information is expected
across Java 8 and 11. This one manifested under JDK 10 and 11 on windows
only; note Jython uses 8 as a compile target.

Stack trace excerpt, see bug 2810 for more detail

```
test_syspath_initializer (test.test_jython_initializer.TestUsingInitializer) ... Exception in thread "main" java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer;
        at org.python.core.io.BufferedReader.clear(BufferedReader.java:147)

```
https://bugs.jython.org/issue2810

Uses static alternatives to the nio.Buffer methods as provided by convenience
classes in backport9. This is a small package, and could be inlined if we don't want to carry another dependency.
https://github.com/headius/backport9

This follows the JRuby solution here https://github.com/jruby/jruby/issues/5450
... though that was running with 8 compiled with 11. Here I am using the jython.exe which is compiled with 11 with an 8 target.

Tested on 8 and 11 on my windows machine with two caveats. 1/ Jeff doesn't see it on his machine 2/ I do not have a good explanation why BufferedReader.clear() causes this problem but eg `ConsoleOutputReader.getPrompt()` does not. As listed in the bug entry, they have very similar bytecode. JIT debug options under openjdk gave me no extra insight, though I'm also also not terribly expert down in the assembly level.

For gradle, I ran the build, but don't have a great way to test jython-slim yet. I wanted the regr test passing first.

Am making this patch available for review as it's a nasty error when it happens, the fix is a low risk one, and people may just have more insight on why it manifests inconsistently.


jeff5:
Do we agree this only happens if you compile on 9+ (using ``-target 1.8``) and run on 8-?

If we want to build using the later compiler (why?), then we're probably doing the wrong thing in using ``-target 1.8``, and we ought to use ``--release 8``. I hadn't looked that up until you pointed out the JRuby ticket.

This didn't completely work for JRuby, but it seems to work for us. It is provided in Ant as an option to ``javac`` (had to update Ant to 1.9.14). Not sure about Gradle, although I did notice Gradle nag me along these lines, maybe when building on 11.

adamburkegh:
Hey replying here as the more recent update but I did see your note on the bug report as well.

> Do we agree this only happens if you compile on 9+ (using -target 1.8) and run on 8-?

Yes, with a but.

I can reproduce it with a very restricted path only including JDK 11, but when I make the culprit test call out to java -version, it reports 1.8.0_181-b13. So there is something environmentally weird.

I can get exactly the condition you describe by compiling on 9+ and running on 8-. That will fail immediately - you can't even get a console - because `PyFile` is called indirectly from `jython.main()`

```
C:\Users\Adam\jython\jython5\jython>dist\bin\jython.exe
Exception in thread "main" java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer;
        at org.python.core.io.BufferedReader.clear(BufferedReader.java:147)
        at org.python.core.io.BufferedReader.<init>(BufferedReader.java:27)
        at org.python.core.PyFile.createBuffer(PyFile.java:227)
        at org.python.core.PyFile.file___init__(PyFile.java:185)
        at org.python.core.PyFile.file___init__(PyFile.java:178)
```

I think it is worth being more defensive. I will give release vs target a go too. 

For my environment this is a corner case that shows when jython calls itself in test_jython_initializer. I can get it even when I set the PATH explicitly to include only JDK11. (I am still puzzling that out). But it is still fundamentally an issue of 8 running against 11-compiled-target-8.

adamburkegh:
Added one more commit which was the main cause. Jython picks up java from java.home before PATH, so was running the spawned Jython using a Java set in a global default JAVA_HOME.

This is not a critical bug, and was mainly caused by a tangle in my environment, but I'm probably not the only one with lots of java versions hanging around on their machine, so I don't think the extra defensiveness hurts.

To summarize, there are three fixes, any one of which independently works:
1/ Use the static / cast idiom to work with Buffer classes (in this patch)
2/ Export JAVA_HOME to ensure consistent use of java version from test_jython_initializer (in this patch)
3/ Change build.xml to use release=8 instead of target=1.8

I tested (3) locally and it worked by itself. However using release doesn't seem to allow different source and target versions, so since we recently said we were keeping source=1.7 for now, I didn't include it.

(1) might be seen as overkill, though I personally prefer the defensiveness. (2) I think is good to include regardless. I can see it either way and am happy to tweak the patch accordingly.

adamburkegh:
As a slight tangent, is there a conventional way to access properties from Jython code, in a way that considers the registry? I scanned the codebase but couldn't see a good example. Our registry doco page doesn't say either. Using `System.getProperty()` in this patch didn't seem ideal.

jeff5:
Ah, in the subprocess: that's a helpful explanation. In my environment I'm always swapping Java and I have a script that does a thorough (?) job. No time to think now but just here to say thanks for clarifying the cause(s).

jeff5:
Normally `System.getProperties() == sys.registry` and although the API seems designed to allow them to be separate I'm not sure when that happens, so when `sys.registry` is to be preferred. Given `java.home` should represent your JVM, I can't see either why you'd set it from Python.

The "bug" here, you point out, is really a weirdness in your environment whereby you run Jython locally overriding the Java version while the sub-process picks up a system-defined version. And this only affects someone who has a Jython built against an API later than the system version. There is some merit in coping with that (or at least 

What you propose with `JAVA_HOME` fixes it for the test. Seeing this, I was wondering why it doesn't fail the same way in all the tests involving a subprocess. I think that is because the environment (therefore `JAVA_HOME`) is normally passed on wholesale. Let's say this test intends not to, it still seems reasonable to pass on that one element.

The JRuby solution is intrusive and we don't have the same need to do it.

But I think `--release 8` is the right and sufficient answer (actually the `javac` property). Done correctly in Ant, the user still has a simple way to override it to 7 (or even 11, I think). What I did as an experiment was to insert
 ```
<property name="jython.java.version" value="8" />
```
(not 1.8) and then in every `javac`
```
                release="${jdk.target.version}"
```
Different property names, but the first feeds the second. Ant now grumbles `Ignoring source, target and bootclasspath as release has been set` so you're right we have lost our built in reminder to maintain source compatibility, but it's only for convenience. Maybe do
```
ant "-Djython.java.version=7"
```
now and then.

adamburkegh:
Looks like gradle doesn't support `release` cleanly yet (and yet ant does which makes me smile ruefully)

https://github.com/gradle/gradle/issues/2510

> `ant "-Djython.java.version=7"` now and then

I never trust my ability to remember this without an automated check : )

More seriously, I think we should just clearly and publicly drop Java 7 source compatibility at 2.7.3 / 3.x, after the official 2.7.2 release.

adamburkegh:
I have pushed a new version of this PR which just ensures JAVA_HOME is set in the corresponding test. It makes it simpler and less error-prone to develop on windows, whether its called a bug or a tweak or whichever.

Javac release=8 change worked, but it seems to make sense to bundle with an actual version change. 

There is a number of new reg test failures on windows coming from trunk. From the chat on the dev list I guess you know about them, but it is better to keep this change separate while the regression pack isn't clean for me locally.

jeff5:
Now in at https://hg.python.org/jython/rev/6d3659465010 . Thanks again.

PR #161, opened 2019-11-15
Expose fork and exec* functions.

hellozyemlya:
os.py check for existence of this functions, but they are not exposed from posix module. Make them exposed to ensure that os module looks more like default "native" os module.

### DIFF
--- a/src/org/python/modules/posix/PosixModule.java
+++ b/src/org/python/modules/posix/PosixModule.java
@@ -130,7 +130,11 @@ public class PosixModule implements ClassDictInit {
         try {
             nativePosix = posix.isNative();
             dict.__setitem__("_native_posix", Py.newBoolean(nativePosix));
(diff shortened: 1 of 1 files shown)


jeff5:
Thanks for tying this. Travis shows a failure, but I think it is unconnected with the patch.

However, when it is applied locally, I get this when I try to use `execv` at the prompt (works in CPython 3):
```
>>> os.execv("C:\\Windows\\notepad.exe", ["test.tmp"])
NotImplementedError: egid
```

In Linux it seems to work, but my terminal is messed up (no echo).

There are no tests, which may be a matter of merging in from CPython stdlib. (What happens here is that, while a lot of tests come from `lib-python`, when we need to tweak one, we fork it into our `~/Lib`, but then they fossilise.)

jeff5:

PR #187, opened 2020-05-25
Travis CI: Test on more JDKs

cclauss:


jeff5:
Thanks @cclauss . After a period of CI being mostly green, the last few commits are all red, due to the osx Java 8 job simply not running at all. I didn't consider that a release blocker because it ran and passed before, and all I changed was the version number.

All jobs ran for you, but some haven't passed. (The osx Java 8 one ran and passed. :)) Maybe we haven't found our limits, or there's some trivial limitation we can avoid. However, I don't want us to add failing CI jobs.

We're in the thick of moving the official project repo to GitHub. I'll be moving issues and PRs across, so the number will change, but maybe we'll get to run the CI again.

cclauss:
The other possibility is to drop Travis CI and replace it with GitHub Actions.
https://github.com/actions/setup-java
https://medium.com/faun/continuous-integration-of-java-project-with-github-actions-7a8a0e8246ef

jeff5:
Unfortunately a PR (unlike an issue) cannot be transferred to the new official home, so this (when we're sure it passes CI) will need to go via a patch.

cclauss:
Let’s close this and I can rebuild it in the new repo.

jeff5:
Ok. Would prefer to transfer issues before anyone creates new issues or PRs, just so the numbers are easy to understand.
Have still to re-enable CI there.

cclauss:
I could create both GitHub Actions CI and Travis CI on the new repo and we can run them thru a side-by-side evaluation. Once we know what we like, we could shut one off and leave the other on.

